PR #84, opened 2019-05-08
Feat notebook

chryswoods:
I've ported the notebook.biosimspace.org docker image over to using the conda-installed BioSimSpace and have updated everything accordingly. This allowed me to use a vanilla jupyterhub docker image, which has simplified things a lot. I've upgraded to the latest stable version of jupyterhub and have tested the image locally. I have also updated the oracle k8s config and have now deployed this to notebook.biosimspace.org (still waiting for the new IP address and SSL cert to propagate - you can see this by connecting directly to http://132.145.251.170 - note that we don't need /hub/tmplogin any more).

I've put git and nano on the notebook service, and cloned biosimspace into the home directory. I've also added a symbolic link to pull "demo" into the home directory to make this easier to find. I've also cloned in the python_and_data workshop, and added a "update_biosimspace" command that will "git pull" and then "python setup.py install" biosimspace live, in case we need to apply last minute fixes.

@lohedges - how do you want to upload the data for your biosimspace workshop? Do you want this in the biosimspace main repo, or as a separate biosimspace_workshop repo that we can clone into the home directory? Please take a look at this. We need to spend some time in person on Monday going through so that we both have knowledge of how to deploy everything.


docker/notebook/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+# pin the version of conda so that we don't get unexpected
+# changes in python
+RUN echo "python 3.7.*" >> /opt/conda/conda-meta/pinned
+
+# Do all conda work as $NB_USER
+USER $NB_USER
+WORKDIR $HOME
+
+# Install Sire via conda
+RUN conda install -c conda-forge -c omnia -c michellab/label/dev sire
```
lohedges: No need to install Sire separately. A compatible version will be pulled in with the BioSimSpace install.

docker/notebook/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+
+# Install BioSimSpace via conda
+RUN conda install -c conda-forge -c omnia -c michellab/label/dev biosimspace
```
lohedges: The latest devel has RDKit as a dependency, which is used to generate pre-matches for the MCS search. The updated Conda install is:

```
conda install -c rdkit -c conda-forge -c omnia -c michellab/label/dev biosimspace
```

docker/notebook/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+
+RUN conda install watchdog
+RUN pip install --upgrade pip && \
+    pip install pygtail && \
```
lohedges: The following pip install line should pull in all non-conda BioSimSpace depencencies:

```sh
pip install fileupload pygtail pypdb
```

docker/notebook/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+    pip install matplotlib && \
+    pip install jupyterhub-tmpauthenticator
+
+# Install and enable nglview (including patching some issues) and fileupload
+COPY nglview_patch.sh $HOME
+RUN pip --no-cache-dir install nglview && \
```
lohedges: NGLView is now a Conda depenency of the BIoSimSpace Conda package, so you no longer need to install via pip. You still need to activate it though, since this is not reliably automated by Conda.

docker/notebook/Dockerfile
```diff
@@ -0,0 +1,99 @@
...
+    jupyter-nbextension install nglview --py --sys-prefix && \
+    jupyter-nbextension enable nglview --py --sys-prefix && \
+    $HOME/nglview_patch.sh /opt/conda && \
```
lohedges: The latest NGLView version no longer requires patching.

chryswoods:
Thanks - please make the changes needed to the Dockerfile and I will rebuild and repush to the notebook server. Please also add in the changes you want for the BioSimSpace workshop

chryswoods:
(my comment is in a weird place as my GitHub thinks that your comments were made 5 hours in the future. I know you are good, but time travel is truly impressive :-))

lohedges:
Okay, it was indeed a channel priority issue, although it didn't cause problems when I tested the package build locally. A working Conda install command is:

```
conda install -c rdkit -c conda-forge -c omnia -c michellab/label/dev biosimspace
```

Also, I tested NGLView and it hasn't been activated properly during the default Conda install, so let's stick with the manual approach that you use in the Dockerfile.

lohedges:
Thanks. GitHub appear to have worked out their server timing issues so I'm no longer a visitor from the future.

I've added the changes to the Dockerfile and removed the redundant patch file. I also modified the `update_biosimspace` script to pull in the latest changes via Conda, i.e. doing a `conda upate`. I'm not sure if I can use the `conda` command directly here? Feel free to change if a path prefix needs to be added.

chryswoods:
Thanks. I'll rebuild and redeploy now. Happily letsencrypt has sorted itself out so notebook.biosimspace.org is now fully back to life.

Are you sure you want to use conda for "update_biosimspace"? I used "git pull" and "setup.py install" as it makes it much easier for you to make quick fixes during a running workshop without having to go to the trouble of building another conda package.

chryswoods:
Everything is now deployed and working at notebook.biosimspace.org

lohedges:
Ah, good point. I had though about keeping dependencies in sync and whatnot, rather than on-the-fly updates during a workshop. I doubt there would be a situation where the list of dependencies changed during a session, so the git pull and regular python install seems fine.

Just realised that I forgot to add RDKit as a depedency for the binary BioSimSpace package. I'll update that in devel, then merge these changes in.

Cheers.

PR #104, opened 2019-08-06
Fixed typo and replaced os.system() calls with native Python calls

ptosco:
- added a missing parenthesis
- replaced os.system() calls with native Python calls

### DIFF
--- a/nodes/playground/prepareFEP.py
+++ b/nodes/playground/prepareFEP.py
@@ -12,6 +12,7 @@
 
 
 import os
+import zipfile
 from Sire.Mol import AtomIdx
 import BioSimSpace as BSS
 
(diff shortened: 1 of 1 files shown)


lohedges:
Thanks. Once merged, I'll update the equivalent notebook so that it matches.

ptosco:
Ah, right, that's what I should have updated in the first place! Will do it myself next time, thanks.

lohedges:
No problem. The notebook was actually corrupted for some reason, so I had to revert to an older version before editing.

My normal approach is to:

* Update the notebook with any edits.
* Test that the notebook works.
* Restart the notebook and clear all the cells. (This will remove metadata so the user has a blank notebook.)
* Export as a regular Python script.
* Save and commit the files.

Since changes to the the `nodes/playground` directory shouldn't trigger a CI build, I also use the git hook mentioned [here](https://biosimspace.org/development.html#committing).

PR #107, opened 2019-08-08
Get the unit tests to pass on Windows

ptosco:
A few changes needed to get the unit tests to pass on Windows.
- The Windows shell does not understand single quotes, only double quotes
- Windows does not support symlinks, so instead of calling the `analyse_freenrg` and `somd` symlinks we need to call `sire_python.exe` and pass the script as first argument

### DIFF
--- a/python/BioSimSpace/FreeEnergy/_free_energy.py
+++ b/python/BioSimSpace/FreeEnergy/_free_energy.py
@@ -31,11 +31,13 @@ __all__ = ["FreeEnergy"]
 from collections import OrderedDict as _OrderedDict
 
 import math as _math
+import sys as _sys
 import os as _os
 import subprocess as _subprocess
 import tempfile as _tempfile
 
 from Sire.Base import getBinDir as _getBinDir
+from Sire.Base import getShareDir as _getShareDir
 
 import Sire.IO as _SireIO
 import Sire.Mol as _SireMol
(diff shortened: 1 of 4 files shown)


lohedges:
Fantastic. Thanks for this. I assume similar changes will be needed to get other processes to work on WIndows too, but we can deal with those as they arise. (It is good to get our bundled software to work properly at a minimum!)

lohedges:
Out of interest, how do you get `Sire.Base.Process`, which is used to handle the molecular simulation processes, to work on Windows? This is the current state of [process_windows.cpp](https://github.com/michellab/Sire/blob/devel/corelib/src/libs/SireBase/process_windows.cpp):

```cpp
#include "sire_process.h"
#include "SireError/errors.h"

using namespace SireBase;

Process::Process()
{
    throw SireError::incomplete_code( QObject::tr(
        "SireBase::Process has not yet been ported to Windows..."), CODELOC );
}

...
```

Do you have your own custom implementation?

ptosco:
@lohedges I am about to submit a PR with my WIndows implementation of `SireBase::Process` plus other critical fixes.

lohedges:
Ah great, I look forward to that.

I was also meaning to talk to you about the possibility of getting your help with adding a Windows build for Sire to our Azure Pipeline. It should then be possible to make a Windows Conda package, following similar logic to what I used for Linux and macOS [here](https://github.com/michellab/Sire/tree/devel/docker/sire-conda). (I've got a `bld.bat` file for the Conda package, but don't have enough PowerShell expertise to figure out the rest of the changes that would be required.)

ptosco:
That would be great. I have a working `compile_sire.bat` which does the same as `compile_sire.sh`, i.e. build Sire from scratch downloading all required packages from `conda` and building `libcpuid` which is not provided by `conda`. I will upload it as part of the PR.

Issue #99, opened 2019-07-31
Mapping/merging errors


dlukauskis:
BSS seems to have issues with growing/breaking three membered rings, as the ligand pair here (13a_17g):

![13a_17g](https://user-images.githubusercontent.com/50554130/62219869-fb165a80-b3a6-11e9-8c2a-793fdc337385.png)

fail to merge, giving this error:
```
Traceback (most recent call last):
  File "prepareFEP.py", line 185, in <module>
    merged = BSS.Align.merge(lig1, lig2, mapping, allow_ring_breaking=node.getInput("allow_ring_breaking"))
  File "/home/e620070/anaconda3/envs/test_BSS/lib/python3.7/site-packages/BioSimSpace/Align/_align.py", line 576, in merge
    property_map0=property_map0, property_map1=property_map1)
  File "/home/e620070/anaconda3/envs/test_BSS/lib/python3.7/site-packages/BioSimSpace/_SireWrappers/_molecule.py", line 2443, in _merge
    raise _IncompatibleError("Inconsistent number of bonds in merged molecule!")
BioSimSpace._Exceptions._exceptions.IncompatibleError: Inconsistent number of bonds in merged molecule!
 ```
There are more error from this particular testset, but lets tackle them one at a time. [Here](https://github.com/michellab/BioSimSpace/files/3452308/Bace_mapping_tests.zip)
 is the folder with structures, mappings and a notebook. In this particular case I used prepareFEP, with --allow_ring_breaking. BSS version 2019.1.0+228.gd744721

**Comment by lohedges** [comment]
Thanks, I'll have a look at our ring opening/closing detection code to see what's going wrong.

**Comment by lohedges** [comment]
I see the problem. At present, when the connectivity between atoms changes, I'm testing whether both atoms are part of a ring in one state, and at least one isn't in the other. Looking at atoms 39 and 20 above, the change goes from one being in the ring (2) and one being _on_ the ring (39), to both not being in the ring. I need to add a check for an atom being _on_ a ring as well as _in_ a ring.

**Comment by lohedges** [comment]
Could you elaborate on what you mean by:

> 3-mem ring issue, mostly

I've made some tweaks to handle the case above and can now merge the ligand pair successfully. However, others in your input set still don't work:

```
CAT-13c CAT-17i
CAT-13d CAT-13f
CAT-13d CAT-13i
CAT-13e CAT-17g
CAT-13e CAT-17i
CAT-13g CAT-17g
[...]

PR #108, opened 2019-08-09
BSS considers rings to be broken also when they are not

ptosco:
The changes introduced to address #99 disallow adding a whole ring where there was none, which should be allowed.
Currently this transformation is disallowed:
```
                        CH2-CH2
0   1         0   1    /       \
CH3-CH3 ----> CH3-CH2-CH       CH2
                       \       /
                        CH2-CH2
```
as atom 1 is not adjacent to a ring in state 1 while it is in state 2, and as a consequence `_is_ring_broken()` returns `True`. However,  no ring is broken as a whole new ring is added in this case.
Also, the `_onRing function` would return false for atom 0 in this case:
```
      CH2-CH2     CH2-CH2
     /       \    /      \
CH3-CH     0 CH-CH       CH2
     \       /    \      /
      CH2-CH2     CH2-CH2
```
because as atom 0 is in a ring, in addition to be adjacent to a ring, so _onRing() would return immediately `False` due to this early `if` clause:
```
    # The atom is in a ring.
    if conn.inRing(idx):
        return False
```
So in such a case the whole ring addition transformation would be allowed because `_is_ring_broken()` would return `False`, which is inconsistent.
Removing the `onRing()` check altogether does not seem to break any unit test.
@lohedges Does the above make sense to you?


lohedges:
This makes complete sense, but I get the following test failures here:

```
test/Align/test_align.py::test_prematch[prematch0] PASSED                                                                                              [  7%]
test/Align/test_align.py::test_prematch[prematch1] PASSED                                                                                              [ 14%]
test/Align/test_align.py::test_prematch[prematch2] PASSED                                                                                              [ 21%]
test/Align/test_align.py::test_prematch[prematch3] PASSED                                                                                              [ 28%]
test/Align/test_align.py::test_invalid_prematch[prematch0] PASSED                                                                                      [ 35%]
test/Align/test_align.py::test_invalid_prematch[prematch1] PASSED                                                                                      [ 42%]
test/Align/test_align.py::test_invalid_prematch[prematch2] PASSED                                                                                      [ 50%]
test/Align/test_align.py::test_invalid_prematch[prematch3] PASSED                                                                                      [ 57%]
test/Align/test_align.py::test_merge PASSED                                                                                                            [ 64%]
test/Align/test_align.py::test_ring_breaking_three_membered FAILED                                                                                     [ 71%]
test/Align/test_align.py::test_ring_breaking_five_membered PASSED                                                                                      [ 78%]
test/Align/test_align.py::test_ring_breaking_six_membered PASSED                                                                                       [ 85%]
test/Align/test_align.py::test_ring_size_change[ligands0] FAILED                                                                                       [ 92%]
test/Align/test_align.py::test_ring_size_change[ligands1] PASSED                                                                                       [100%]
```

The tests that failed were why the `onRing` logic was introduced. I'll have a think about this, since clearly I didn't consider the ring growth example that you show.

ptosco:
Ah, OK, I thought I should have broken something! Let me have a 2nd look then.

lohedges:
No problem. I think this is a difficult problem and there are clearly a lot of edge cases to address. I was basically using whatever information that I could extract from the Sire `Connectivity` object, but it might be required that we supplement the information somehow. I also think that the order in which checks are done (to eliminate various situations) will be quite important.

lohedges:
Do you have some example input files for this so that I can debug locally and write some tests?

ptosco:
[growring.tar.gz](http://1b3211c549dbdbb4dbb9-1f72b29fc48de44749ce23b81e3e52fc.r32.cf3.rackcdn.com/growring.tar.gz)

lohedges:
Great. I'll take a look this afternoon.

ptosco:
@lohedges This seems to get all tests to pass (including the new one).

lohedges:
Many thanks for this, and for the unit test too. I'll check it against the other problem mappings from a few of the other issue threads and merge when I'm happy.

lohedges:
The changes don't cause any issues with any of the other problem mappings that had previously been fixed. There are still issues with the recent mappings mentioned in [this](https://github.com/michellab/BioSimSpace/issues/99) thread, but I think those are the result of different bugs and can be debugged independently.

I'm happy that we're moving in the right direction. Thanks again for catching this, and for the elegant fix.

ptosco:
Thanks a lot!

lohedges:
Just to say that the ligand pair at the bottom of [this](https://github.com/michellab/BioSimSpace/issues/99) issue (the `p38` pair) appears to have been missed by the ring-breaking/growing checks. I'm working on catching this perturbation but let me know if you have any suggestions.

PR #110, opened 2019-08-12
Avoid that test_somd.py fails due to OpenMM nan issues

ptosco:
If test_somd.py fails due to a `RuntimeError: Particle coordinate is nan` error, it should pass with a warning rather than failing, as it is not really a problem with BSS or SOMD.

### DIFF
--- a/test/Process/test_somd.py
+++ b/test/Process/test_somd.py
@@ -1,3 +1,5 @@
+import os
+import warnings as _warnings
 import BioSimSpace as BSS
 
 def test_minimise():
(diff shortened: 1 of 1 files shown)


ptosco:
@lohedges I noticed that occasionally `test_somd.py` fails due to OpenMM issues with minimising coordinates. This should not be flagged as as a BSS failure (as I myself initially thought when I first saw the failure), as it is not really BSS, SOMD or SIre's fault and it may just happen randomly.

lohedges:
Thanks. I agree that this is a random failure and should be flagged accordingly. Do you have any idea why you are seeing it on your system, since it hasn't yet occurred in any of the Azure Pipeline builds (hundreds of builds on Linux and macOS), or in any of my local tests. Perhaps the version of OpenMM that you are using is different?

ptosco:
I have seen this on Windows using OpenMM 7.3.0. To trigger it I ran 1000 times the tests and have only seen it a few times. I will try doing the same on Linux to see if it is a Windows-only thing.

ptosco:
OK, I have reproduced this also on Linux, using OpenMM 7.3.0.
```
for i in `seq 1000 2000`; do
    sire_python -m pytest -v test/Process/test_somd.py > test_somd_${i}.log 2>&1
done
```
After ~150 runs I checked the logs and found one `nan` failure:
```
grep -iw nan test_somd_????.log
test_somd_1043.log:  /build/common/cresset/BioSimSpace/test/Process/test_somd.py:62: RuntimeWarning: Test raised RuntimeError: Particle coordinate is nan
test_somd_1043.log:    _warnings.warn("Test raised RuntimeError: Particle coordinate is nan", RuntimeWarning)
```

lohedges:
Thanks for taking the time to check this. I guess I have just been very lucky this far :-)

ptosco:
I'll check if using OpenMM 7.3.1 makes a difference and get back to you.

Issue #121, opened 2019-09-19
Peturbation files peturb between atom types that should be the same


mark-cresset:
# Expected Behavior

Creating a peturbation file between two molecules; one has a tetrahydropyran and the other a piperidine. Both molecules are otherwise identical, and both have a thiophene elsewhere in their structure.

On parameterising the molecules with teLeap, the carbon atoms in the thiophene are assigned either cc or cd atom types. It appears to be a bit random which atom type is given to which atom: in particular for this pair atoms C4 and C5 are assigned cc/cd in one molecule and cd/cc in the other.

As these atoms are identical in both molecules I would not expect any parameter change for them. 

# Current Behavior

The current behaviour is that peturbation entries are written transforming atom C4 from type cc to type cd, and atom C5 from cd to cc. The LJ parameters are the same, but the bond lengths and angles are not. As a result, there is a parameter change in this part of the molecule despite there being no structural change.

Paolo suggests that changes between supposedly-equivalent AMBER types (cc/cd, ce/cf. etc) be ignored in creating .pert files.
[somd-pert-bug.zip](https://github.com/michellab/BioSimSpace/files/3630930/somd-pert-bug.zip)

## Steps to Reproduce

somd-freenrg.py -C Config.txt

The input files are in the attached tarball.

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System: Linux Debian stable
* Installation method: Flare

lohedges:
I'm not an expert in this, but are the differences in, e.g. bond lengths, between the two ends states a result of the charge distribution of the molecule? (Structurally the substructure of the two end states is the same, but the charge is distributed differently across the whole molecule.) Perhaps @jmichel80 has a better idea about this. I'd be happy to implement the changes proposed by @ptosco if it makes physical sense.

lohedges:
Could you upload Mol2/PDB files for the two molecules so that I can run the parameterisation, mapping, and merge myself?

jmichel80:
I also need the initial input to dig in, but it sounds like it could be an atom typing perception error from parmchk. Could be because of differences in the initial coordinates of the thiophene ring atoms.

lohedges:
Hi @mark-cresset, any update on this? I'd be happy to try to debug if I had access to the original Mol2/PDB files.

lohedges:
[...]

PR #140, opened 2020-02-13
Ensure consistent output across different runs with same input

ptosco:
Currently running BioSimSpace multiple times on the same input yields completely different outputs (seen on `prmtop` and `inpcrd` files).
This happens because the Sire `MolGroupsBase::molecules()` method returns molecules by iterating on a `QHash`, so they are not ordered as in the input file, but based on a hash.
Using the `MolGroupsBase::molecules(const MGID &mgid)` method ensures that molecules are output in the same order as in the input.
While this does not change the outcome of a calculation in any way or form, having consistent outputs is important for 1) tidiness 2) spotting out differences/mistakes.

Also sections in the `.pert` file are currently output based on a Python `dict` ordering rather than being consistently sorted based on same criterion. Again, this does not change results but having consistent outputs is important for 1) tidiness 2) spotting out differences/mistakes.

### DIFF
--- a/python/BioSimSpace/_SireWrappers/_molecule.py
+++ b/python/BioSimSpace/_SireWrappers/_molecule.py
@@ -991,8 +991,20 @@ class Molecule(_SireWrapper):
 
             # First create records for the bonds that are unique to lambda = 0 and 1.
 
+            def sort_bonds(bonds, idx):
+                # Get the bond potential.
+                bond = bonds[idx]
+
+                # Get the AtomIdx for the atoms in the bond.
+                idx0 = info.atomIdx(bond.atom0())
+                idx1 = info.atomIdx(bond.atom1())
+                
+                return (mol.atom(idx0).name().value(),
+                        mol.atom(idx1).name().value())
+
             # lambda = 0.
-            for idx in bonds0_unique_idx.values():
+            for idx in sorted(bonds0_unique_idx.values(),
+                              key=lambda idx: sort_bonds(bonds0, idx)):
                 # Get the bond potential.
                 bond = bonds0[idx]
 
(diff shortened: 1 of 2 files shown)


lohedges:
Thanks @ptosco, I'll test this locally then accept when I'm happy that things pass. I totally agree with you on this. I had taken the same approach when extracting molecules from a `System` object, i.e. they are always returned in the same order, but hadn't applied this to the extraction of potential terms when constructing the pert file.

(I think there were also issues with certain file format conversions where the molecule/atom order wasn't preserved, mostly because of the parallel nature of the parser. I believe that I have caught all of this, but might be wrong.)

lohedges:
The sorting helper functions, e.g. `sort_bonds`, return a tuple of atom names. I guess this works because I enforce unique atom names at the start of the `_toPertFile` function. This is required since the pert file uses names, rather than indices, to distinguish atoms. In Sire there is no reason why names have to be unique, so you could, for example, end up with multiple bonds between two atoms called `C`. (The names would be the same, but the indices are unique.)

I note that, although the ordering is still preserved, the atom naming in the pert and prm files will still be different due to the random suffix that I apply to duplicate names. This was a simple catch all solution that I implemented that adheres to the AMBER fixed width naming convention. Are you okay with this, or would you prefer if the naming was preserved too? I'm sure I can come up with something simple if needed.

Cheers.

ptosco:
Many thanks @lohedges for the quick reply. Happy to keep it as is, at the moment I don't mind if names in the `.pert` files are different from those in the `.prmtop` files.
The purpose of my PR was more to get consistency in `.pert` and `.prmtop` outputs across different runs with the same topology, but different coordinates.
At the moment it is quite hard to detect the actual identity across runs which are supposed to only vary in coordinates, as the `.pert` and `.prmtop` files, though computationally equivalent, are indeed quite different due to the unstable sorting of molecules and potential terms.

ptosco:
Thanks @lohedges, that was blazing fast!

lohedges:
No problem. Out of interest, do you know if [this](https://github.com/michellab/BioSimSpace/issues/121) issue was ever resolved? I never managed to get an appropriate set of input files to test things locally. Ideally I'd need prm/rst files and a mapping to test properly.

ptosco:
I'll check with Mark. I don;t have those input files either, unfortunately.

PR #142, opened 2020-02-14
Integrates PR #140

ptosco:
This PR introduces a few more stable sorts that I had missed in #140.


lohedges:
Thanks @ptosco, all good this end so happy to accept.

ptosco:
@lohedges Thanks!

PR #173, opened 2021-01-07
Add support for exporting CWL wrappers from BioSimSpace nodes

lohedges:
This pull request adds functionality to export [Common Workflow Language](https://www.commonwl.org/) (CWL) wrappers from a BioSimSpace node, for example:

```python
python minimisation.py --export-cwl
```

This writes a wrapper called `minimisation.cwl` in the current directory, which can be run as part of CWL workflow as follows:

```bash
cwltool minimisation.cwl config.yaml
```

Depending on your setup, you might need to run:

```bash
cwltool --preserve-entire-environment minimisation.cwl config.yaml
```

Here `config.yaml` is a CWL YAML configuration file, e.g.:

```yaml
files:
   - {class: File, path: /home/lester/BioSimSpace/demo/amber/ala/ala.top}
   - {class: File, path: /home/lester/BioSimSpace/demo/amber/ala/ala.crd}

steps:
  1000
```

At present, using BioSimSpace within CWL is limited to the use of `BioSimSpace.Gateway.File` and `BioSimSpace.Gateway.FileSet` output requirements, which cover the majority of use cases. Due to the way in which CWL works, the prefix used for output files must match the name used for the requirement, e.g. if a requirement was called `output`, then a file might be named `output.txt`. This allows the use of `glob` in the CWL `outputBinding` functionality. This requirement is automatically enforced so that files will be renamed when a mismatch is found.

Any unit based input requirement, e.g. `BioSimSpace.Gateway.Length`, should be specified as a CWL `string` type to allow for greatest flexibility, e.g.:

```yaml
length:
  25 Angstroms
```

(Floats also work, but they don't allow the user to specify the requirement using a different unit or notation. All BioSimSpace requirement types can be implicitly converted from a string.)

Unfortunately CWL currently doesn't have support for placing restrictions/constraints on the input values. See [here](https://github.com/common-workflow-language/common-workflow-language/issues/764) for a discussion. This means that other tools can't use a CWL wrapper to generate handy things like widgets and dropdowns for specifying input requirements, like those used when a BioSimSpace node is run from within Jupyter. This can't be difficult to add, since it was easy to set up in BioSimSpace. I imagine it's just a case of agreeing on a scope/standard. There does seem to be some recent activity on the above thread, so hopefully it's something that's in the pipeline. (I don't see how it's possible to create useful KNIME nodes from a CWL wrapper unless this feature is added.)

### DIFF
--- a/doc/source/nodes.rst
+++ b/doc/source/nodes.rst
@@ -159,6 +159,17 @@ Here ``config.yaml`` is a YAML configuration file, e.g.:
     steps:
     1000
 
+At present, using BioSimSpace within CWL is limited to the use of
+:class:`BioSimSpace.Gateway.File <BioSimSpace.Gateway.File>`
+and
+:class:`BioSimSpace.Gateway.FileSet <BioSimSpace.Gateway.FileSet>`
+requirements, which cover the majority of use cases. Due to the way in which
+CWL works, the prefix used for output files must match the name used for the
+requirement, e.g. if a requirement was called ``output``, then a file might be
+named ``output.txt``. This allows the use of ``glob`` in the CWL ``outputBinding``
+functionality. This requirement is automatically enforced so that files will
+be renamed when a mismatch is found.
+
 Forwards compatibility
 ======================
 
--- a/doc/source/nodes.rst
+++ b/doc/source/nodes.rst
@@ -134,9 +134,9 @@ function.)
 Common Workflow Language
 ========================
 
-It is also possible to export a node as a Common Workflow Language
-`(CFL) <https://www.commonwl.org/>`__ wrapper. For example, using the
-``minimisation.py`` example from the previous section:
+It is also possible to export a node as a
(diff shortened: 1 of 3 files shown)


mr-c:
Hello, I just discovered https://biosimspace.org/nodes.html#common-workflow-language ; super cool! Would you be interested in presenting your experience in exporting to CWL at an upcoming CWL video chat?

We video chat every Tuesday at 3:30 PM UK summer time / 4:30 PM central european summer time

lohedges:
Hi there. I'd be happy to give a quick overview. There's not too much to add, other than what's in the documentation and pull request already, but it would be good to know if I've missed anything obvious, or if there are any new features in the pipeline, e.g. constraints on input values.

We wanted to support CWL as a potential way of interfacing with external workflow  managers in the future, without the headache of maintaining code for each tool. I don't think anyone has used our wrappers in production, but we do have a few simple examples that I could walk through.

Cheers.

mr-c:
@lohedges Great! Can you email me? mrc@commonwl.org

PR #181, opened 2021-01-27
OpenFF integration improvements

lohedges:
Following yesterday's discussions, I have implemented @j-wags and @mattwthompson's recommendation to create an OpenFF Molecule using an intermediate SDF file written by RDKit, since this quietly handles some tricky sanitisation issues.

I've also added some fixes to correctly expand verbose error messages when the parameterisation process is run as a background thread.

### DIFF
--- a/python/BioSimSpace/Parameters/Protocol/_amber.py
+++ b/python/BioSimSpace/Parameters/Protocol/_amber.py
@@ -372,6 +372,7 @@ class GAFF(_protocol.Protocol):
         except Exception as e:
             msg = "Failed to write system to 'PDB' format."
             if _isVerbose():
+                msg += ": " + getattr(e, "message", repr(e))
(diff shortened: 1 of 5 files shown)


mattwthompson:
This passes an "eye test" from me; Jeff may have more to say. Are there any tests/use cases to also check, or other feedback I could provide?

lohedges:
Thanks. I agree that it would be good to add a few tests, although at this stage it would be good to push a devel build out to our collaborators so that I can get feedback on what particular systems they are struggling with. Are there particular molecules that you'd recommend? It would be good to compile a small set of PDBs that we know should work. Perhaps including a few that only work reliably using the SDF intermediate.

Cheers.

lohedges:
Merging into `devel` for now so that users can experiment. Will push any further improvements to the feature branch which can be merged across again in future.

Issue #194, opened 2021-03-02
[TUTORIAL] Funnel metadynamics


lohedges:
This is a thread to discuss the creation of a tutorial showing how to implement funnel metadynamics within BioSimSpac.(Tagging @dlukauskis.)

jmichel80:
Adding current diagram summarising the use case 

![image](https://user-images.githubusercontent.com/11722212/109876826-22f82f00-7c6a-11eb-8329-46973149a197.png)

dlukauskis:
I've ran a fun-metaD sim that was setup for gromacs and I got a few errors.

1. Missing definition for the ligand center of mass, after the definition of p1 and p2 points.
```lig: COM ATOMS=first_ligand_atom_id-last_ligand_atom_id```

2. Missing sigma parameter for the extent CV, so instead of 
``` metad: METAD ARG=pp.proj,pp.ext SIGMA=0.025 ...```

    it should be

    ``` metad: METAD ARG=pp.proj,pp.ext SIGMA=0.025,0.05 ...```

lohedges:
Thanks for catching these. I'll be able to push a fix tomorrow.

dlukauskis:
I've made some functions that draw the funnel for NGLview, here's [the code](https://github.com/michellab/BioSimSpace/files/6184859/nglview_funnel.zip). The idea behind using NGLview is to rapidly check that the funnel points out the right way and includes/excludes the right bits of the protein. @lohedges, could you have a go at making my functions built into BSS code?

lohedges:
Thanks, @dlukauskis. I've fixed the issues with the GROMACS PLUMED file and have added a check so that the ProjectionOnAxis.cpp file is only copied to the working directory when the PLUMED version is less than 2.7.

The plotting functionality looks great. I'll start implementing this now. Just to check... you define three vectors: `origin`, `v1`, and `v2`. Both `origin` and `v1` are identical in this case. Which system did you use to define the values of `v1` and `v2`? Are these just the COM of `p0` and `p1` used to define the funnel CV? (From reading the function doc strings, it would seem so.) I see you include a `solvated.pdb` file. Does this correspond to one of the systems from your original tutorial?

dlukauskis:
Oh yeah, origin = v1 = p1 and v2 = p2. This was based on some older bit of code so I forgot to remove all of the redundancies like this. Solvated.pdb is 2WI3, which is almost identical to 2WI2 from the tutorial, except for a slightly different binding pose.

[...]

PR #201, opened 2021-04-17
OMM Device index bug fix

dlukauskis:
Fixes a bug that looks like this 
```
File "/home/dom/anaconda3/envs/bss/lib/python3.7/site-packages/simtk/openmm/app/simulation.py", line 105, in __init__
             self.context = mm.Context(self.system, self.integrator, platform, platformProperties)
         File "/home/dom/anaconda3/envs/bss/lib/python3.7/site-packages/simtk/openmm/openmm.py", line 18607, in __init__
             this = _openmm.new_Context(*args)
         NotImplementedError: Wrong number or type of arguments for overloaded function 'new_Context'.
         Possible C/C++ prototypes are:
             OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &)
             OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &,OpenMM::Platform &)
         OpenMM::Context::Context(OpenMM::System const &,OpenMM::Integrator &,OpenMM::Platform &,std::map< std::string,std::string,std::less< std::string >,std::allocator< std::pair< std::string const,std
::string > > > const &)
         OpenMM::Context::Context(OpenMM::Context const &)
```
It basically doesn't support  `properties = {CudaDeviceIndex: cuda_devices}`, when I select CUDA as the platform. Simply giving `properties = {}` fixes the issue. Relevant to #194

### DIFF
--- a/python/BioSimSpace/Process/_openmm.py
+++ b/python/BioSimSpace/Process/_openmm.py
@@ -1496,22 +1496,7 @@ class OpenMM(_process.Process):
         # Set the simulation platform.
         self.addToConfig("\n# Set the simulation platform.")
         self.addToConfig(f"platform = Platform.getPlatformByName('{self._platform}')")
-        if self._platform == "CPU":
-            self.addToConfig("properties = {}")
-        elif self._platform == "CUDA":
-            cuda_devices = _os.environ.get("CUDA_VISIBLE_DEVICES")
-            if cuda_devices is None:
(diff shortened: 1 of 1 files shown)


lohedges:
The issue was that I was trying to set a property formatted as a dictionary inside of at Python f-string. I've used regular string formatting and it should now work. Could you test the update and let me know if it still doesn't work. Perhaps it's still more reliable to just leave the properties blank (as you did) and let OpenMM autodetect things.

(A new conda package will be available in about an hour or so.)

Cheers.

PR #202, opened 2021-04-17
Device indices

dlukauskis:
I tested your new commit for this issue and it raised the same error OpenMM as I showed in the previous PR. The device indices need to be strings, not integers.


dlukauskis:
I didn't see that you put the quotation marks around the device index, so your changes were correct all along.

lohedges:
Ah, sorry. I immediately reallised my mistake after I pushed the first commit and pushed another a few minutes later. You must have only pulled the first change.

PR #205, opened 2021-04-21
Small typo + more informative reporter

dlukauskis:
Relevant for #194


lohedges:
Thanks for this. I've added to the feature branch that I'm currently working on and this will be closed when I merge across into devel after testing.

With regards to the state reporter: As default, I just added what matched the other engines that we support so that we can get consistent information by calling, e.g. `process.getPotentialEnergy()`, etc. I could wrap the extra options that you've added, but those won't be available across all engines. (We do have some other engine-specific reporters, though.) For now I'll just leave it as extra information that an experienced user could monitor by looking at the log file directly, or by running `process.stdout()`.

Cheers.

PR #211, opened 2021-05-12
Feature feed lomap scores

JenkeScheen:
Adjusted `_Align.py` such that it can now take in a variable `scores_file` (will throw an exception when `names` is not given simultaneously). This file contains user-input edge scorings that LOMAP can use instead of the native LOMAP score. `_Align.py` contains some checks to see if all possible combinations of `names` are in `scores_file`; the ordering should not matter for LOMAP because `lomap_networkgen.py` parses the file per perturbation.

I built the `--ml_pd` parameter and subsequent `ml_pd` path string passing in a similar way to how the fingerprint (`-f`) is passed to `compute_mtx()`.

Bonus: adjusted `generateNetwork()` such that now it uses a .mol2 intermediate instead of .pdb by naming molecules as {index}_{name}.mol2. With some small changes in the rdkit preparation, now pydot plots molecules with bond orders and non-polar hydrogens removed.

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -171,14 +171,27 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
 
         # check if all the molecule transformations are in the passed file.
         perturbations = _itertools.combinations(names, 2)
-        with open(scores_file) as scores_file_:
-          contents = repr(scores_file_.read())
+        with open(scores_file, "r") as scores_file_:
+          contents = []
+          for line in scores_file_:
+            # check that str,str,float. Because of how file is parsed, 
+            # lig1_name and lig2_name are always str.
+            lig1_name, lig2_name, score = line.rstrip().split(",")
+            try: 
+              float(score)
+            except ValueError:
+              raise ValueError(f"{scores_file} contains a non-numerical value on third "\
+                                +f"column ({line.rstrip()}). Make sure that each line "\
+                                +"in the file is formatted as str,str,value.")
+            contents.append(line.rstrip())
 
+          # check that all possible perturbations in the ligand series are accounted for in
+          # the scores_file.
           for pert in perturbations:
-            if not f"{pert[0]},{pert[1]}" in contents:
+            if not f"{pert[0]},{pert[1]}" in ",".join(contents):
 
               # check the opposite direction as well.
-              if not f"{pert[1]},{pert[0]}" in contents:
+              if not f"{pert[1]},{pert[0]}" in ",".join(contents):
 
                 raise ValueError(f"Could not find {pert[0]},{pert[1]} (or the inverse) "+\
                   f"in {scores_file}. Make sure your input file contains all possible transformations.")
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -38,6 +38,7 @@ import os as _os
 import subprocess as _subprocess
 import sys as _sys
 import tempfile as _tempfile
+import itertools
 
 import warnings as _warnings
 # Suppress numpy warnings from RDKit import.
@@ -76,7 +77,7 @@ except:
     _fkcombu_exe = None
 
 def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
-        input_scores=None, property_map={}):
+        scores_file=None, property_map={}):
     """Generate a perturbation network using Lead Optimisation Mappper (LOMAP).
 
        Parameters
@@ -97,10 +98,13 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
            If using a 'work_dir', then a PNG image will be located in
            'work_dir/images/network.png'.
 
-       input_scores : [float]
-           A list of float values (0.0-0.1) for the molecules. If defined, LOMAP will 
-           use these values to score molecular transformations instead of using its
-           own LOMAP score.
+       scores_file : str
+           Path to a CSV file in the form:
+           lig1,lig2,score
+           ..
+           lig_ lig_,score
+           , where score is a float that is fed into LOMAP for edge scoring
+           (instead of the default LOMAP score based on MCSS).
 
        property_map : dict
            A dictionary that maps "properties" in molecule0 to their user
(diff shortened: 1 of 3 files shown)


JenkeScheen:
I see now that commit 91ac7c9 is old and builds on my initial idea of passing an array of `input_scores` to `generateNetwork()`; please disregard this code as it is removed in commit 6912e0c (i.e. where they are passed as file).

lohedges:
Thanks, @JenkeScheen, this looks great. I'll take a look and update with any comments / suggestions.

python/BioSimSpace/Align/_align.py
```diff
@@ -38,6 +38,7 @@
 import subprocess as _subprocess
 import sys as _sys
 import tempfile as _tempfile
+import itertools
```
lohedges, line 41: As a convention, all external modules are listed alphabetically and imported privately so that they aren't exposed directly to the user, even if they import `_align.py` directly, e.g:

```python
import itertools as _itertools
```

This makes it obvious what  functionality comes from BioSImSpace when you import anything, i.e. only the top-level API of BioSimSpace is exposed, no private functions, external modules, variables, etc.

python/BioSimSpace/Align/_align.py
```diff
@@ -97,6 +98,14 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
            If using a 'work_dir', then a PNG image will be located in
            'work_dir/images/network.png'.
 
+       scores_file : str
+           Path to a CSV file in the form:
+           lig1,lig2,score
+           ..
+           lig_ lig_,score
+           , where score is a float that is fed into LOMAP for edge scoring
+           (instead of the default LOMAP score based on MCSS).
```
lohedges, line 107: Can we say "Maximum Common Substructure (MCS)" here, just in case the user doesn't know what the acronym means. (Unlikely, but it's good to be explicit.)

python/BioSimSpace/Align/_align.py
```diff
@@ -145,9 +154,32 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
...
     if type(property_map) is not dict:
         raise TypeError("'property_map' must be of type 'dict'")
 
+    # Validate the scores file.
+    if scores_file is not None:
+      if type(scores_file) is not str:
+          raise TypeError("'input_scores' must be of type 'str'")
+      if names is None:
+          raise ValueError("'names' must be defined when passing 'scores_file' to LOMAP.")
+
```
lohedges, line 167: It would be good to validate that the file exists before opening it so that we can report a useful message to the user, e.g.:

```python
if _os.path.isfile(scores_file):
    # Do stuff
    ...
else:
    raise IOError(f"The scores file didn't exist: {scores_file}")
```

(The contextmanager would raise an exception for us, but it's nice to make it a little clearer.)

python/BioSimSpace/Align/_align.py
```diff
@@ -168,17 +200,26 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
     _os.makedirs(work_dir + "/outputs", exist_ok=True)
 
     # Write all of the molecules to disk.
-    for x, molecule in enumerate(molecules):
-        _IO.saveMolecules(work_dir + f"/inputs/{x:03d}",
-            molecule, "pdb", property_map=property_map)
+    for x, (molecule, name) in enumerate(zip(molecules, names)):
```
lohedges, line 203: Yes, I think that Mol2 will be more robust in general. Glad the visualisation now works with Mol2 as well. I kept getting blank images when I tried it for some reason.

I do worry about conversions to Mol2 from other formats, e.g. if the original molecule was a PDB file. I think I'll update this after the merge so that we preferentially use Mol2, but write to PDB if that was the original format. (In this case, Mol2 bond-orders would be missing anyway.)

python/BioSimSpace/Align/_lomap/lomap_networkgen.py
```diff
@@ -530,11 +535,37 @@ def ecr(mol_i, mol_j):
                         logging.info('MCS molecules: %s - %s' % (self[i].getName(), self[j].getName()))
 
                     if not fingerprint:
-                        # Maximum Common Subgraph (MCS) calculation
-                        MC = mcs.MCS(moli, molj, options=self.options)
-                        ml=MC.all_atom_match_list()
-                        self.set_MCSmap(i,j,ml)
-                        MCS_map[(i,j)]=ml
+                        if ml_pd:
```
lohedges, line 538: Do we want to pre-validate the contents of the scoring file in anyway? I know that we check that all ligand pairs are present, but would it also be useful to check that each line contains `str`, `str`, `float` when it is split on commas? Might be useful to do at the start of the `generateNetwork` function where the ligand names are validated, just to catch any surprises. It looks like you would just get  a `KeyError` here if the value was missing. I'm not sure what would happen if the value was a string rather than a float, for example.

python/BioSimSpace/Align/_align.py
JenkeScheen: agreed, I think in its current form nothing would break but the user would just not see bond orders when porting PDB files. IMO there should be a user-warning saying it's generally a bad idea to use PDB files for ligand structures.

python/BioSimSpace/Align/_lomap/lomap_networkgen.py
JenkeScheen: I think it would be good to validate the contents in some way because it takes a while to trace LOMAP errors back when the input is off.

I'll build that check into `generateNetwork()`; the validation code block is getting quite extensive though. Might need some clean-up at some point.

python/BioSimSpace/Align/_align.py
JenkeScheen: see f93dfa096da5bb4be09259edeab90a2f1f77c767

python/BioSimSpace/Align/_align.py
JenkeScheen: see f93dfa096da5bb4be09259edeab90a2f1f77c767

python/BioSimSpace/Align/_align.py
JenkeScheen: see e69de1fc4b19487ce19247ae83369b34d1a3e419

python/BioSimSpace/Align/_lomap/lomap_networkgen.py
JenkeScheen: yes, I think the check should be in `_Align.py`, because inputting a str,str,str line in the `scores_file` makes LOMAP fail silently because it just excludes that particular MCS:
`lomap.out`:
```
Skipping MCS molecules (exception): 009_ejm_43.mol2 - 012_ejm_42.mol2	

could not convert string to float:[str]
```

JenkeScheen: see 35138541aa20ea2a9781a95cd868fe24cc4d660d. I allowed users to input ints as well; LOMAP has no problem with scores >1 so in this way it would be fine if a user inputs `scores_file` where scores are any of [0,1,2,..n].

JenkeScheen: I think if you agree with the changes in this one, all should be resolved. Thanks for the feedback!

lohedges:
> good one! I think I'll go for the second idea, wouldn't hurt to leave the option open for future dev. If at some point users want to pass more values to LOMAP then that could easily be processed in `compute_mtx()`.

Great, thanks! Adding new features to BioSimSpace essentially involves trying to think of all the ways that a user might break it, as well as all the ways that someone might want to use it in future.

I'll merge this across now.

Issue #165, opened 2020-11-24
Feature discussion: multi-step perturbation


jmichel80:
Opening this thread here for discussions with @JenkeScheen . The current focus is SOMD and initial tests will be done using simple python scripts but it would be useful to eventually support this from within BioSimSpace to facilitate integration with existing workflows and to roll out functionality to other free energy codes. 

In some cases it may be desirable to implement a multi-staged lambda schedule where only some terms of the potential energy function are coupled to a lambda parameter at a given time. For instance, a general scheme that could apply to a morph that contains both atoms that need to become dummy atoms ('to dummy'), or start as dummy atoms ('from dummy'), or atoms that remain present throughout ('hard atoms') could be implemented using 6 steps described below.

![image](https://user-images.githubusercontent.com/11722212/100082308-fdc07d00-2e3f-11eb-9386-89db6439c8c4.png)

Each step should have its own pert file. The same morph topology should be applicable throughout. 
This leads to six simulations each with their own lambda schedule to be determined. Each simulation should be analysed separately with analyse_freenrg mbar. The sum of the six resulting free energy changes should be identical (to within uncertainties) to that obtained from the direct path sketched at the top of the figure (in practice this would only be true for well behaved systems). 

Some of the steps could be lumped together for efficiency, e.g. bonded terms + hard atoms charges/LJ if testing reveal these converged rapidly.

JenkeScheen:
Thanks for this Julien. I should be able to get started on writing a script that does this.. Do you think we'd run into trouble with this protocol when trying to transform A to B with mixed growing/shrinking, i.e.:
![image](https://user-images.githubusercontent.com/43140137/100084163-555fe800-2e42-11eb-9dda-ef87f3949175.png)
?

jmichel80:
Should be covered by the example I posted. 
In some schemes (where e.g. you only need to convert some atoms into dummy atoms) you may not need steps 5 and 6.

lohedges:
This should be simple enough to implement in the current BioSImSpace set up. We would just need to add a parameter the the protocol for a single free energy leg that lists the terms that are perturbed (or are excluded from the perturbation if that's simpler). That way you could do something like:

```python
protocol = BSS.Protocol.FreeEnergy(terms="charge")
```

This would write a pert file for the lambda leg that only includes the specified terms.

JenkeScheen:
That would be ideal Lester, would you think it best that in this case we would declare six protocols and set up a directory with each? i.e.
```
[...]

PR #213, opened 2021-05-20
Feature multi step perturbation 3

JenkeScheen:
This PR relates to #165.

This is a first implementation of the multi-step protocol; with this it is now possible for the user to define a `pert_type` in `BSS.Protocol.FreeEnergy()`. If not defined, the setup will default to a normal FEP setup (i.e. 1 step). I've built it into Protocol because it gives users the freedom to define specific numbers of lambda windows per step.

There are several caveats to the current implementation:
- when building multiple steps, `BSS.Protocol.FreeEnergy()` will have to be called multiple times. For instance:
```
# set up step 1.
protocol = BSS.Protocol.FreeEnergy(num_lam=10, pert_type="1_discharge_soft")
BSS.FreeEnergy.Solvation(system_solvated, protocol, work_dir="output/1_discharge_soft")

# set up step 2.
protocol = BSS.Protocol.FreeEnergy(num_lam=15, pert_type="2_vanish_soft")
BSS.FreeEnergy.Solvation(system_solvated, protocol, work_dir="output/2_vanish_soft")

# etc.
```
- Currently for a given perturbation BSS will setup whichever step is specified. When (e.g.) a perturbation only consists of **growing** atoms, two of the five steps are redundant (.pert files of `1_discharge_soft` and `2_vanish_soft` contain no changes). At the moment it is up to the user to take this into account: either by knowing which perturbations are grow/shrink/both and to setup specific steps accordingly, or by removing redundant step folders by e.g. checking if the .pert files contain changes. Implementing something like this in BSS might be quite complex.
- `freenrg.analyse()` will only compute the free energy for a given step and not the sum of steps.

### DIFF
--- a/python/BioSimSpace/Protocol/_free_energy.py
+++ b/python/BioSimSpace/Protocol/_free_energy.py
@@ -89,8 +89,13 @@ class FreeEnergy(_Protocol):
            restart_interval : int
                The frequency at which restart configurations and trajectory
            perturbation_type : str
-                Which type of perturbation to create a multistep free energy work_dir with. 
-                Defaults to standard, i.e. onestep.
+               The type of perturbation to perform. Options are:
+                'standard' : A full perturbation of all terms (default option).
+                '1_discharge_soft' : perturb all discharging soft atom charge terms (i.e. value->0.0).
+                '2_vanish_soft' : perturb all vanishing soft atom LJ terms (i.e. value->0.0).
+                '3_flip' : perturb all hard atom terms as well as bonds/angles.
+                '4_grow_soft' : perturb all growing soft atom LJ terms (i.e. 0.0->value).
+                '5_charge_soft' : perturb all charging soft atom LJ terms (i.e. 0.0->value).
         """
 
         # Call the base class constructor.
--- a/python/BioSimSpace/Protocol/_free_energy.py
+++ b/python/BioSimSpace/Protocol/_free_energy.py
@@ -96,6 +96,8 @@ class FreeEnergy(_Protocol):
                 'flip' : perturb all hard atom terms as well as bonds/angles.
                 'grow_soft' : perturb all growing soft atom LJ terms (i.e. 0.0->value).
                 'charge_soft' : perturb all charging soft atom LJ terms (i.e. 0.0->value).
+
+                Currently GROMACS only supports perturbation_type=full.
         """
 
         # Call the base class constructor.
--- a/python/BioSimSpace/Protocol/_free_energy.py
+++ b/python/BioSimSpace/Protocol/_free_energy.py
@@ -48,7 +48,8 @@ class FreeEnergy(_Protocol):
                  temperature=_Types.Temperature(300, "kelvin"),
                  pressure=_Types.Pressure(1, "atmosphere"),
                  report_interval=100,
-                 restart_interval=500
+                 restart_interval=500,
+                 pert_type="standard"
                 ):
         """Constructor.
 
@@ -87,6 +88,9 @@ class FreeEnergy(_Protocol):
 
            restart_interval : int
                The frequency at which restart configurations and trajectory
+           pert_type : str
+                Which type of perturbation to create a multistep free energy work_dir with. 
+                Defaults to standard, i.e. onestep.
         """
 
         # Call the base class constructor.
@@ -116,6 +120,9 @@ class FreeEnergy(_Protocol):
         # Set the restart interval.
         self.setRestartInterval(restart_interval)
 
+        # Set the perturbation type. Default is "standard", i.e. onestep protocol.
+        self.setPertType(pert_type)
+
     def __str__(self):
         """Return a human readable string representation of the object."""
         if self._is_customised:
@@ -135,6 +142,31 @@ class FreeEnergy(_Protocol):
                     "runtime=%s, temperature=%s, pressure=%s, report_interval=%d, restart_interval=%d)"
                    ) % (self._lambda, self._lambda_vals, self._timestep, self._runtime,
                         self._temperature, self._pressure, self._report_interval, self._restart_interval)
+
+    def getPertType(self):
+        """Get the type of step of the multistep approach to write. Default ("standard") is onestep.
+
+           Returns
+           -------
+
+           pert_type : str
+               The perturbation type.
+        """
+        return self._pert_type
+
+    def setPertType(self, pert_type):
+        """Set the time step.
+
+           Parameters
+           ----------
+
+           pert_type : str
+               The perturbation type.
+        """
+        if type(pert_type) is str:
+            self._pert_type = pert_type
+        else:
+            raise TypeError("'pert_type' must be of type str")
 
     def getLambda(self):
         """Get the value of the perturbation parameter.
--- a/python/BioSimSpace/Protocol/_free_energy.py
+++ b/python/BioSimSpace/Protocol/_free_energy.py
@@ -49,7 +49,7 @@ class FreeEnergy(_Protocol):
                  pressure=_Types.Pressure(1, "atmosphere"),
                  report_interval=100,
                  restart_interval=500,
-                 pert_type="standard"
+                 perturbation_type="standard"
                 ):
         """Constructor.
 
@@ -88,7 +88,7 @@ class FreeEnergy(_Protocol):
 
            restart_interval : int
                The frequency at which restart configurations and trajectory
-           pert_type : str
+           perturbation_type : str
                 Which type of perturbation to create a multistep free energy work_dir with. 
                 Defaults to standard, i.e. onestep.
         """
@@ -121,7 +121,7 @@ class FreeEnergy(_Protocol):
         self.setRestartInterval(restart_interval)
 
         # Set the perturbation type. Default is "standard", i.e. onestep protocol.
-        self.setPertType(pert_type)
+        self.setPertType(perturbation_type)
 
     def __str__(self):
         """Return a human readable string representation of the object."""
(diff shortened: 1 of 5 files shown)


python/BioSimSpace/_SireWrappers/_molecule.py
```diff
@@ -1266,13 +1270,22 @@ def _toPertFile(self, filename="MORPH.pert", zero_dummy_dihedrals=False,
 
         if type(property_map) is not dict:
             raise TypeError("'property_map' must be of type 'dict'")
-
-        # Seed the random number generator so that we get reproducible atom names.
-        # This is helpful when debugging since we can directly compare pert files.
-        _random.seed(42)
```
JenkeScheen, line 1272: I restored this in 956ece1ecabcb09b793d7e4c833b2b63426c2f6b

python/BioSimSpace/_SireWrappers/_molecule.py
```diff
@@ -1242,17 +1243,20 @@ def _toPertFile(self, filename="MORPH.pert", zero_dummy_dihedrals=False,
...
+               approach in perturbing the LJ, bond and angle terms.
+
            Returns
            -------
 
            molecule : Sire.Mol.Molecule
                The molecule with properties corresponding to the lamda = 0 state.
         """
-
-        if not self._is_merged:
```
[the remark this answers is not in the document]
JenkeScheen: FYI, there are several cases below this line with the same issue

JenkeScheen (commented):
Note that I'd made an active decision to keep everything in conditional blocks as IMO the steps of writing these pert files are complicated enough - wrapping steps in functions adds an extra layer of complexity

lohedges:
Thanks, Jenke. I'll take a look this afternoon and post some comments / suggestions when I get a chance.

python/BioSimSpace/_SireWrappers/_molecule.py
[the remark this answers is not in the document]
lohedges: This is because you've changed the name of the `self` argument in the method signature to `molecule`. While `self` is purely convention, it is good to be consistent so that it is obvious that this refers to the object itself, rather than a molecule that was passed in. As such, could you revert things back to how they were.

python/BioSimSpace/Protocol/_free_energy.py
```diff
@@ -48,7 +48,8 @@ def __init__(self,
                  temperature=_Types.Temperature(300, "kelvin"),
                  pressure=_Types.Pressure(1, "atmosphere"),
                  report_interval=100,
-                 restart_interval=500
+                 restart_interval=500,
+                 pert_type="standard"
```
lohedges, line 52: Could this be named `perturbation_type` for clarity with `getPerturbationType` and `setPerturbationType` methods.

python/BioSimSpace/Protocol/_free_energy.py
```diff
@@ -136,6 +143,31 @@ def __repr__(self):
...
+           Parameters
+           ----------
+
+           pert_type : str
+               The perturbation type.
+        """
+        if type(pert_type) is str:
+            self._pert_type = pert_type
+        else:
+            raise TypeError("'pert_type' must be of type str")
```
lohedges: Could you also add the check that you perform in `_SireWrappers/_molecule.py` here, i.e.:

```python
          if type(pert_type) is not str:
              raise TypeError("'pert_type' must be of type 'str'")

          allowed_pert_types = [
                                "standard",
                                "1_discharge_soft",
                                "2_vanish_soft",
                                "3_flip",
                                "4_grow_soft",
                                "5_charge_soft"]
          if pert_type not in allowed_pert_types:
            raise ValueError("'pert_type' must be any of: "+str(allowed_pert_types), pert_type)
```

I would also add `perturbation_type = perturbation_type.lower().replace(" ", "")` to catch case and whitespace errors before checking against the list of allowed values.

It's good to catch errors as soon as they occur, rather than at the point at which the parameter is used. Still leave the check in `_molecule.py`, since that will catch the same errors if a user happens to use the `_toPertFile` method directly.

python/BioSimSpace/_SireWrappers/_molecule.py
```diff
@@ -1269,10 +1273,24 @@ def _toPertFile(self, filename="MORPH.pert", zero_dummy_dihedrals=False,
...
+          if type(pert_type) is not str:
+              raise TypeError("'pert_type' must be of type 'str'")
+
+          allowed_pert_types = [  
+                                "standard",
+                                "1_discharge_soft",
+                                "2_vanish_soft",
+                                "3_flip",
+                                "4_grow_soft",
+                                "5_charge_soft"]
```
lohedges, line 1288: With regards to the numbering: Am I right in remembering that the steps need to be performed in a specific order? If not, it might be better to remove numbering. Also, is it possible to conceive of other perturbation types that could be added in addition to these, e.g. as other steps in between? If so, then we'd need to renumber if we did add support for other perturbation types, which would break backwards compatibility.

python/BioSimSpace/_SireWrappers/_molecule.py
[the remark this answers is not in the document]
JenkeScheen: Ah! I must have changed this during development last winter as I was importing the `toPertFile()` directly. See 91b34392651e4f8b049e4ae844bc426afa595f23

python/BioSimSpace/Protocol/_free_energy.py
JenkeScheen: see ab8be7b04c708df335aa173061156f0415e1be3b

lohedges (changes requested):
Thanks, this looks good. The main thing that is missing is handling of the new pertubation types with GROMACS, i.e. the other FEP engine that we support. At present it would just perform a standard protocol, even when a different perturbation type is chosen, i.e. breaking interoperability.

If this isn't possible to implement with GROMACS then we need to raise an `BioSimSpace._Exeptions.IncompatibleError` in `BioSimSpace.Process._gromacs.py` if `protocol.getPerturbationType()` is anything other than `standard`. It would also be good to do this in `BioSimSpace.FreeEnergy._free_energy.py` too whenever `engine='gromacs'`, so that the error is caught early. If this _could_ be implemented with GROMACS, then we should raise a `NotImplementedError` instead. This is easy for me to add so I'd be happy to do it myself after merging.

With regards to your comment on creating multiple `FreeEnergy` objects, i.e. one for each step, it could be possible to handle situations where a user passes in a _list_ of protocol objects, however, the logic for setting up all of the directories and processes could become complicated. It might just be easier (although less elegant) to set things up step-by-step, as you propose.

python/BioSimSpace/Protocol/_free_energy.py
```diff
@@ -87,6 +88,9 @@ def __init__(self,
 
            restart_interval : int
                The frequency at which restart configurations and trajectory
+           pert_type : str
+                Which type of perturbation to create a multistep free energy work_dir with. 
+                Defaults to standard, i.e. onestep.
```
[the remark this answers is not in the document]
lohedges: We could also mention that the option only applies to FEP simulations with SOMD, if that is the case.

python/BioSimSpace/_SireWrappers/_molecule.py
JenkeScheen: I added the numbering at the time to make the steps somewhat more intuitive (to me, at least). In principle the steps are not dependent on each other in any way, but users will have to run all five of them to get meaningful free energies (by summation).

IIRC `3_flip` consists of multiple actions, i.e. perturbing bonds, angles and all hard atom parameters. If ever we would want to break that up into two steps then yes, we would have to renumber the steps in that case, but I'm not sure why we would want to do that (famous last words)

JenkeScheen:
I'm not sure how/ if this could be implemented in the GROMACS side of things - perhaps @jmichel80 could comment on whether we should include a `NotImplementedError` or an `IncompatibleError`.

jmichel80:
For GROMACS a ''Not implemented'' message will suffice as we currently have no plans to test that this functionality works for GROMACS.

python/BioSimSpace/_SireWrappers/_molecule.py
JenkeScheen: okay, I'll remove the numbering.

JenkeScheen: I'll also adjust `standard` to `full`, I agree this is clearer.

JenkeScheen: see 44cce8823acde7e44a5bf5defdbe1dd822973815 and 2fa10bb54f3315a5d2cf7a44d16224e51d2392d4

python/BioSimSpace/Protocol/_free_energy.py
JenkeScheen: see 2fa10bb54f3315a5d2cf7a44d16224e51d2392d4

python/BioSimSpace/Protocol/_free_energy.py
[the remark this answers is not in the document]
JenkeScheen: see f19873e88cde7d3db1c91ab2cd8de5f529ef4162 and 06718383bd3d196d699541e73a747b3d84834ece

JenkeScheen:
I think everything has been processed except for the `NotImplementedError`. @lohedges could you advise where exactly to implement this?

lohedges:
Thanks, I'll take a closer look at the edits tomorrow. You'll need to add the `NotImplementedError` in two places:

https://github.com/michellab/BioSimSpace/blob/04381784ab8d2e2abacf0466d9cce25fdaabbc57/python/BioSimSpace/Process/_gromacs.py#L161-L167

I'd suggest adding:

```python
if self._protocol.getPerturbationType() != "full":
    raise NotImplementedError("..."")
```

within the first conditional block, i.e. after checking that the system contains a merged molecule. The text should state the only the "full" perturbation type is supported by GROMACS and also suggest that the user uses SOMD if a different perturbation type is required.

It would also be good to add the same check here:

https://github.com/michellab/BioSimSpace/blob/04381784ab8d2e2abacf0466d9cce25fdaabbc57/python/BioSimSpace/FreeEnergy/_free_energy.py#L152-L153

I'd split this into two checks, i.e.:

```python
if engine == "GROMACS":
    # Make sure GROMACS is installed if GROMACS engine is selected.
    if _gmx_exe is None:
        raise _MissingSoftwareError("Cannot use GROMACS engine as GROMACS is not installed!")
    # Make sure the perturbation type is supported.
    if protocol.getPerturbationType() != "full":
        raise NotImplementedError("...")
```

JenkeScheen:
see f866bf2c991524bcf37a06ad129d8edfef384001 for GROMACS checks.

JenkeScheen:
Great! Thanks a lot! Let's discuss when you're ready to update the documentation - I have made some figures/ explanations over the last few months that could help.

PR #224, opened 2021-07-23
Update _io.py

kexul:
small typo fix


lohedges:
Thanks, will merge in now. Think I'll also refactor that section a bit to make it a bit clearer.

PR #236, opened 2021-09-27
small bug fix 

kexul:
`allowed` is not defined, replace it with self.restraints()


lohedges:
Thanks for catching. Forgot to change this when updating the API.

PR #237, opened 2021-10-06
added UpdatePropertyCache() call to resolve depiction bug

JenkeScheen:
this fixes a small bug that I was having while generating FEP networks for the CDK2 benchmarking series of ligands. The UpdatePropertyCache() call takes care of valency issues in the input file.


JenkeScheen:
related to this issue from way back: https://github.com/rdkit/rdkit/issues/149

PR #244, opened 2021-11-04
[WIP] Add AMBER FE integration into BioSimSpace

msuruzhon:
This pull request is related to the integration of free energy calculations using AMBER within BioSimSpace. 

I have made some changes around the `Process` and `Protocol` objects and we should integrate these changes fully within the other engines as well, before we merge to `devel`. Also, I am writing some `FreeEnergyEquilibration` and `FreeEnergyMinimisation` classes, which will need to be pushed and integrated additionally before the merge.

Until then, it might be worth adding a separate `conda` channel, e.g. `michellab/label/amber`, based on this branch. Would something like this be possible? Many thanks.


lohedges:
Thanks so much for this. I should be able to take a look and make comments/suggestions early next week. It should be possible to set up a specific AMBER conda build within the CI, although it will need a bit of tweaking.

Cheers,
Lester

lohedges:
(Just flagging this as a work in progress [WIP] to avoid accidental merge.)

Thanks again for this. At a quick pass things look really good. Thanks for following the style guide closely, it'll make it much easier to review and suggest edits. I'm compiling a list of comments and suggestions as I go and will post some discussion items here along with formatting tweaks / suggestions in the actual code review. (Probably some time next week.)

Cheers.

msuruzhon:
That sounds great, thanks a lot for that Lester. I might push some extra commits in the meantime so hopefully GitHub notifies you when this happens.

lohedges:
Yes, feel free to keep working on things. I've added the Exscientia repo as a remote of mine so I can pull in updates at any time and review/test locally.

msuruzhon:
Hi @lohedges, do you think we should set softcore potentials to be on by default (with some reasonable values)? I am currently refactoring the GROMACS/SOMD parameter generators and this stood out to me. We have to use softcore potentials for AMBER, as they require a certain format of prm7/rst7 files, so I thought it would make sense if we did that for GROMACS and SOMD as well?

Also, I noticed that the default cutoffs were different for SOMD, AMBER and GROMACS. I have currently set all of them to 0.8 nm. Would there be any issues with this?

lohedges:
Hi there, these are good questions:

>  ...do you think we should set softcore potentials to be on by default

Hmm, I'm not sure, I think @jmichel80 would have a better idea. Naively I would assume that we'd want to implement the _best practice_ approach for the given FEP implementation, or the one that is most robust. I agree that it would be good to be consistent across engines, however, since this would allow a more direct comparison of results, but it might turn out that one engine is unreliable or far less optimised with one approach. (Maybe softcore could be an option to the FEP protocol, if possible.)

> Also, I noticed that the default cutoffs were different for SOMD, AMBER and GROMACS. I have currently set all of them to 0.8 nm. Would there be any issues with this?

This is a good question and is something that is hard to solve in general. In practice, the cutoffs should actually be a property of the force fields that are used to parameterise the molecules in the system, rather than something that is (somewhat) arbitrarily set when generating the config. However, only OpenFF provides this information directly so it would be hard to know (without plotting the potential, I guess). When starting on the project I hadn't actually run extensive simulations with all of the engines so essentially aggregated advice on best practice from various tutorials online and from colleagues in the Chemistry department at Bristol. I'd be happy to standardise the cutoffs across engines, assuming there are no peculiarities in the implementation of the specific integrators that mean that the same cutoff doesn't give the same results.

jmichel80:
You should use the best practice settings for each implementation by default. For example for cutoffs, SOMD uses reaction-field and need a larger cutoff than AMBER and GROMACS/PME. Softcore potentials are always on when using SOMD etc...It's fine to question what the default settings for some implementations are as it also helps define what best practice mean for a given implementation. This may also evolve as we interface with new releases over time, which adds another layer of complexity, especially with code that is not clearly versioned.

lohedges:
Thanks, @jmichel80 that pretty much mirrors my thoughts. As the code evolves I think it would be good to have the `Process` objects (or higher level tools like `FreeEnergy.Relative`) to write out specifcs of the implementation that was performed by the engine (stuff that might not be obvious from looking at the protocol or configuration file). This would provide greater transparency as to what exactly is going on under the hood and would make it easier to make comparisons between results.

msuruzhon:
Thanks, @lohedges and @jmichel80 for the replies. Does SOMD use Coulomb softcore by default then? If so, which parameters does it use, as I think it would make sense to put them as GROMACS and AMBER defaults as well.

In terms of the cutoffs, should we keep the GROMACS/AMBER defaults at 0.8 nm, while the SOMD ones at 1 nm then? I think that would make sense.

jmichel80:
hi @msuruzhon see page S-2 of https://pubs.acs.org/doi/suppl/10.1021/acs.jctc.8b00544/suppl_file/ct8b00544_si_001.pdf for details about the softcore implementations (may have been deprecated by newer AMBER and Gromacs releases). We found that different code needed different parameters due to slight differences in the implementations. For RBFE with SOMD I recommend n=0, delta=1.0 (alpha in the notation used in the linked pdf).

msuruzhon:
Thanks @jmichel80 I have decided against adding more GROMACS defaults for now, as their Coulomb softcore implementation enforces the same parameter as the LJ one and this results in practically unstable behaviour. Therefore, one needs to detect the direction of perturbation, so that the alchemical protocol is spllit in the correct direction. This could be something that we implement, but right now I am sticking to the previous `BioSimSpace` defaults, which don't use any softcore.

I have largely added all the contributions I have intended to make so I think at this point we should make sure the code is not buggy and is integrated well into `BioSimSpace`.

lohedges:
Thanks, this is looking great. I'll comment on specifics of the code when I get a chance to review but here are a few points of discussion to raise. These mostly concern interoperability and what features we want to expose at this stage:

_ConfigFactory_:

This works very nicely, great job. It solves the issue of customising the protocol at the time the process is instantiated, whereas at present you can only tweak things afterwards (which would be a pain for an entire set of FEP simulations, for example). I had originally tried something like this, but got frustrated at the different way in which "configurations" were expressed. It's tricky when things aren't always clean key/value type options, when a configuration option is conditional on the value of another, and also when information can be contained in multiple places, e.g. with GROMACS. In this case we can assume that anyone who is choosing to override the defaults knows what they are doing, so we can't provide any (or much) help when things go wrong.

At present the `ConfigFactory` is directly exposed to the user through the public API. (It's imported in the `__init__.py` of the `Protocol` sub-package.) I'd prefer if it was only used internally for now, at least until we have ported all engines over. I'd also prefer if it was isolated in an internal  `_Config` sub-package with a `Config` base class along with derived `Config.Amber`, `Config.Gromacs`, etc classes. This will make it easier to test and to add functionality for additional engines.

The new code has removed the use of the `property map` when checking system properties. While currently not widely used, this is a fundamental feature in Sire so we still need any new code to be consistent. For example, the property map allows a user to completely re-map the names of system/molecule properties, so we need to extract them using the same naming scheme, e.g. we need things like

```python
space = system._sire_object.property(property_map.get("space", "space"))
```

The protocol/configuration is probably the most contentious part of BioSimSpace, i.e. what is _best practice_ and it is great to make it easier to customise the options. The original code involved a lot of duplication since I decided to isolate the configuration for an engine specific protocol entirely to allow a user to read through it line-by-line to see what was going on. At this stage it makes much more sense to modularise things for flexibility.

_Additional MCS options_:

I'm not sure if we want to add support for _dual_ and _hybrid_ mappings without having a clear way of supporting these within BioSimSpace (at least interoperably). Since the mapping is just a Python dictionary, there would be no way to know that the user generated a non single-topology MCS, which could then be used for a merge and to run an incorrect single-topology FEP simulation. Perhaps we just need to better highlight that BioSimSpace currently only supports (interoperably) single-topology simulations?

Another issues is the flakiness of the MCS engines. We currently use Sire as a fallback engine when RDKit fails to find the user pre-match in the mapping. In my experience this occurs surprisingly often since, without altering the internals of RDKit itself, the pre-match check can only be applied _after_ the MCS is generated, i.e. it is only ever applied to the _largest_ mapping, even if a smaller one _does_ include the pre-match. In contrast, Sire can validate the pre-match at each stage of the MCS so it will return the largest match including the pre-match. However, Sire doesn't support all of the RDKit options, so these have to be disabled, which would be the case for the additional topology options too.

_Hydrogen mass adjustment_:

As above, the `setHydrogenMass` functionality also doesn't have a clear use within BioSimSpace. (I assume it's being called from your own scripts.) Over the weekend I had a quick play around and have implemented full hydrogen mass repartitioning. (See [this](https://github.com/michellab/BioSimSpace/tree/feature_hmr) branch.) I'd rather this was used instead, since the use is clear. You can now do:

```python
system.repartitionHydrogenMass(mass=4)
```
This repartitions the mass of all hydrogen atoms, reducing the mass of all heavy atoms to which they are bonded. (@jmichel80 I currently don't have any checks for a lower bound on heavy atom masses, let me know if you have suggestions for this, or I can try to follow what is done in SOMD.)

_Free-energy specific protocols_:

I see that there are now additional, free-energy specific, versions of existing protocols, e.g. `FreeEnergyMinimisation`. I've not checked in detail, but how do these differ to the existing implementations? Do they allow you to minimise/equilibrate at a specific intermediate value of lambda? At present, _all_ of the existing protocols can be run using systems containing perturbable molecules with _all_ engines. This works by extracting the lambda = 0 state and performing some tweaks to enable simulation of non-zero mass dummy atoms with the various engines. When `.getSystem()` is called on a process, the updated coordinates are then copied back into original system. (If desired, you can also simulate the lambda = 1 state.) This approach has been tested with SOMD and GROMACS and seems to work well.  We've also used the same approach for minimising/equilibrating with  `pmemd.cuda`, since it was much quicker in our FEP workflow pipeline. (An alternative approach that has been used is to simply minimise and equilibrate the two protein + ligand complexes independently, then perform the merge afterwards.) 

I'd rather not have a proliferation of protocols if the AMBER specific implementation could be handled in the same way. If it did require additional information from the free energy protocol, e.g. lambda, then perhaps this could be added as an additional option to the existing protocols. Alternatively, would it be possible to support your new approach with AMBER and GROMACS too? SOMD does have its own mimimisation and equlibration (equilibrate to lambda) routines, but these aren't decoupled from the full FEP code.

I guess this opens up a discussion as to what is the best way to prepare a system for an FEP simulation. Maybe this could be tested before deciding on which implementation we go with. Ideally it would be nice if there was one obvious approach that was exposed to the user.

Cheers.

(I'll update if I think of anything else to discuss.)

jmichel80:
hi, to add to this. I don't think we can fully decouple free energy setup (MCS options...) from the chosen execution engine in general. I think the API should allow that for power users, but there should be a high level abstraction (e.g. AmberFEPsim, SOMDFEPsim) that makes the correct setup decisions given an execution engine specified ahead of time.  Interoperability between FEP engines would then be handled by starting from the specification of the end states.

lohedges:
Okay, but at present the FEP setup (perturbable system preparation) and simulation are fully decoupled, i.e. you start by creating a merged molecule with no knowledge of the resulting simulation engine that will be used for FEP. Such a change would take a lot of refactoring and probably involve breaking changes. Currently BioSimSpace is assuming that the `System` and `Protocol` are the base objects which are independent of the underlying simulation engine. If the MCS was specific to the engine, then this would no longer be true, i.e. the `System` would now be coupled to an engine.

msuruzhon:
> Thanks, this is looking great. I'll comment on specifics of the code when I get a chance to review but here are a few points of discussion to raise. These mostly concern interoperability and what features we want to expose at this stage:
> 
> _ConfigFactory_:
> 
> This works very nicely, great job. It solves the issue of customising the protocol at the time the process is instantiated, whereas at present you can only tweak things afterwards (which would be a pain for an entire set of FEP simulations, for example). I had originally tried something like this, but got frustrated at the different way in which "configurations" were expressed. It's tricky when things aren't always clean key/value type options, when a configuration option is conditional on the value of another, and also when information can be contained in multiple places, e.g. with GROMACS. In this case we can assume that anyone who is choosing to override the defaults knows what they are doing, so we can't provide any (or much) help when things go wrong.

Yes, that was my intention too. I think the best model is to have some sensible defaults, but also the ability to override as many of them as possible. Since many of these options don't have counterparts in other MD engines, it only makes sense that the user supplies the desired parameters with their MD-engine-specific names.

> 
> At present the `ConfigFactory` is directly exposed to the user through the public API. (It's imported in the `__init__.py` of the `Protocol` sub-package.) I'd prefer if it was only used internally for now, at least until we have ported all engines over. I'd also prefer if it was isolated in an internal `_Config` sub-package with a `Config` base class along with derived `Config.Amber`, `Config.Gromacs`, etc classes. This will make it easier to test and to add functionality for additional engines.
> 

This sounds good, I wasn't sure where to put it when I first created it and I just left it there. I think it makes sense to restructure it in this way.

> The new code has removed the use of the `property map` when checking system properties. While currently not widely used, this is a fundamental feature in Sire so we still need any new code to be consistent. For example, the property map allows a user to completely re-map the names of system/molecule properties, so we need to extract them using the same naming scheme, e.g. we need things like
> 
> ```python
> space = system._sire_object.property(property_map.get("space", "space"))
> ```
> 

I wasn't sure which functions need to have these `property_map`s, so I didn't include them for fear of breaking something. Do you only put them in functions where you call `_sire_object.property()`?

> The protocol/configuration is probably the most contentious part of BioSimSpace, i.e. what is _best practice_ and it is great to make it easier to customise the options. The original code involved a lot of duplication since I decided to isolate the configuration for an engine specific protocol entirely to allow a user to read through it line-by-line to see what was going on. At this stage it makes much more sense to modularise things for flexibility.

Agreed.

> _Additional MCS options_:
> 
> I'm not sure if we want to add support for _dual_ and _hybrid_ mappings without having a clear way of supporting these within BioSimSpace (at least interoperably). Since the mapping is just a Python dictionary, there would be no way to know that the user generated a non single-topology MCS, which could then be used for a merge and to run an incorrect single-topology FEP simulation. Perhaps we just need to better highlight that BioSimSpace currently only supports (interoperably) single-topology simulations?

As @jmichel80 said, it would be nice to have the option of having more control over the `RDKit` parameters. As with the custom configuration parameters, it has to be assumed that the user knows what they are doing, but I guess that's why we put it as an optional argument in the first place. Maybe we can be more clear in the documentation about where one can use these options? The good news is that the single topology MCS is good for all of the engines so far, so it is the obvious default.

> Another issues is the flakiness of the MCS engines. We currently use Sire as a fallback engine when RDKit fails to find the user pre-match in the mapping. In my experience this occurs surprisingly often since, without altering the internals of RDKit itself, the pre-match check can only be applied _after_ the MCS is generated, i.e. it is only ever applied to the _largest_ mapping, even if a smaller one _does_ include the pre-match. In contrast, Sire can validate the pre-match at each stage of the MCS so it will return the largest match including the pre-match. However, Sire doesn't support all of the RDKit options, so these have to be disabled, which would be the case for the additional topology options too.

I think the best way to handle this is to just output a warning or an error if this happens while the topology has been set to something else than the default?

> _Hydrogen mass adjustment_:
> 
> As above, the `setHydrogenMass` functionality also doesn't have a clear use within BioSimSpace. (I assume it's being called from your own scripts.) Over the weekend I had a quick play around and have implemented full hydrogen mass repartitioning. (See [this](https://github.com/michellab/BioSimSpace/tree/feature_hmr) branch.) I'd rather this was used instead, since the use is clear. You can now do:
> 
> ```python
> system.repartitionHydrogenMass(mass=4)
> ```
> 
> This repartitions the mass of all hydrogen atoms, reducing the mass of all heavy atoms to which they are bonded. (@jmichel80 I currently don't have any checks for a lower bound on heavy atom masses, let me know if you have suggestions for this, or I can try to follow what is done in SOMD.)

This is very nice, thanks.

> _Free-energy specific protocols_:
> 
> I see that there are now additional, free-energy specific, versions of existing protocols, e.g. `FreeEnergyMinimisation`. I've not checked in detail, but how do these differ to the existing implementations? Do they allow you to minimise/equilibrate at a specific intermediate value of lambda? At present, _all_ of the existing protocols can be run using systems containing perturbable molecules with _all_ engines. This works by extracting the lambda = 0 state and performing some tweaks to enable simulation of non-zero mass dummy atoms with the various engines. When `.getSystem()` is called on a process, the updated coordinates are then copied back into original system. (If desired, you can also simulate the lambda = 1 state.) This approach has been tested with SOMD and GROMACS and seems to work well. We've also used the same approach for minimising/equilibrating with `pmemd.cuda`, since it was much quicker in our FEP workflow pipeline. (An alternative approach that has been used is to simply minimise and equilibrate the two protein + ligand complexes independently, then perform the merge afterwards.)

It would be nice to have control over the specific lambda value too, since an equilibrated structure at one lambda value does not imply that it is a reasonable structure for another lambda value. That's why I created these new protocols, since they are a generalisation of the regular MD ones. They also make the config generation much easier because by inheriting from two classes, you are automatically activating two separate parts in the config code that add more options relevant to the protocol.

> I'd rather not have a proliferation of protocols if the AMBER specific implementation could be handled in the same way. If it did require additional information from the free energy protocol, e.g. lambda, then perhaps this could be added as an additional option to the existing protocols. Alternatively, would it be possible to support your new approach with AMBER and GROMACS too? SOMD does have its own mimimisation and equlibration (equilibrate to lambda) routines, but these aren't decoupled from the full FEP code.

I think if we have separate `Production` and `FreeEnergy` protocols, it makes sense to have the equivalent for the `Minimisation` and the `Equilibration` ones. Either way, I do not particularly mind not including `FreeEnergyMinimisation` and `FreeEnergyEquilibration`, as they are very simple implementations and one can just as easily define them in a custom code. However, I think that it still makes sense to define the `FreeEnergy` protocol through a mixin, so that one can create these protocols modularly if needed without the need to copy-paste chunks of code.

> I guess this opens up a discussion as to what is the best way to prepare a system for an FEP simulation. Maybe this could be tested before deciding on which implementation we go with. Ideally it would be nice if there was one obvious approach that was exposed to the user.

As with the previous points, I am personally happy with any defaults of your choice, I just prefer having the means to change some of the parameters relatively painlessly if needed. Maybe we could discuss this point further.

Many thanks for the detailed review. How are we going to proceed with the changes? Would it make sense to have a meeting at some point to discuss this further?

lohedges:
Great, thanks for the clarifications.

> I wasn't sure which functions need to have these property_maps, so I didn't include them for fear of breaking something. Do you only put them in functions where you call _sire_object.property()?

For the purposes of generating the configurations, this is pretty much the case. However, there are many Sire related functions that also take a property map as an argument. (For example, a function might use atom masses so would need to know if they have been renamed.) When I go through the code I can point out where anything is missing.

> Maybe we can be more clear in the documentation about where one can use these options?

Yes, I think as long as things are well documented it should be fine. As you say, these are optional arguments and in most cases people will be using the defaults, or calling the `merge` function directly thus bypassing `matchAtoms` entirely.

> I think the best way to handle this is to just output a warning or an error if this happens while the topology has been set to something else than the default?

Yes, this is what is done for the other RDKit specific options that are ignored when Sire is used as a fallback. See [here](https://github.com/michellab/BioSimSpace/blob/4f37a274b1346674307b626a5d06e2907c0393cd/python/BioSimSpace/Align/_align.py#L679-L683) for details.

> It would be nice to have control over the specific lambda value too, since an equilibrated structure at one lambda value does not imply that it is a reasonable structure for another lambda value. That's why I created these new protocols, since they are a generalisation of the regular MD ones. They also make the config generation much easier because by inheriting from two classes, you are automatically activating two separate parts in the config code that add more options relevant to the protocol.
...
I think if we have separate Production and FreeEnergy protocols, it makes sense to have the equivalent for the Minimisation and the Equilibration ones. Either way, I do not particularly mind not including FreeEnergyMinimisation and FreeEnergyEquilibration, as they are very simple implementations and one can just as easily define them in a custom code. However, I think that it still makes sense to define the FreeEnergy protocol through a mixin, so that one can create these protocols modularly if needed without the need to copy-paste chunks of code.

Yes, I definitely agree with this and I like the modularity that the new approach provides. I think in practice @jmichel80 has found that (for SOMD at least) it was easier to equilibrate within the FEP production run itself, i.e. just discarding the initial unequilibrated part of each lambda window when it comes to performing the analysis. If it's possible to implement lambda specific protocols for the other engines then I'd certainly prefer to use your approach. I imagine this should be possible with GROMACS at least. For SOMD I'll have a think about how to extract the minimisation and equilibration routines from the main code so that they can be run independently. (They can be run already, but I don't think they return a trajectory or report statistics in the same way as one would currently expect, so wouldn't be consistent with the existing protocols.) I'm also not too sure about the robustness of SOMD's "equlibrate-to-lambda" code too. Initially we thought about providing a means of generating intermediate lambda states with BioSimSpace (so we could extract them to use in regular protocols) but these are somewhat dependent on the engine in question.

> Many thanks for the detailed review. How are we going to proceed with the changes? Would it make sense to have a meeting at some point to discuss this further?

Yes, it would probably make sense to have another meeting once I've had a chance to go through things more closely and suggest edits.

jmichel80:
Hi @msuruzhon , @annamherz is having a go at testing AMBER FEP calculations with your code changes. Would you be able to share example input files and binding free energy estimate for a representative pair of compounds ? We would like to run things via our FEP interface and check whether the results are consistent with those from other engines, and those you obtain with your local implementation.

msuruzhon:
Hi @jmichel80, thanks for that, I am sending some example input files for binding and solvation free energies taken from [here](https://gitlab.com/RutgersLBSR/alchemical_fe/-/tree/master): [FE_validation.zip](https://github.com/michellab/BioSimSpace/files/7504222/FE_validation.zip)

It makes sense to start with some simple solvation free energies first, such as ethane\~methane, ethane\~methanol and methanol\~methane. As for the binding free energies, we could do ejm42\~ejm54, ejm42\~ejm55 and ejm55\~ejm54.

I haven't run these yet so I don't have any free energy estimates but it might be useful to do these separately and make sure that all engines output the same results.

annamherz:
Hello! I've been running tests for the RBFE w the tyk2 ligands, so here is an update for those! The overall scripts I used I adapted from the BioSimSpace tutorials execution model (https://github.com/michellab/BioSimSpaceTutorials), together with template files I had from previous runs for the preparation at each lambda window (point 3). There were some issues w the run crashing and the ligand leaving the pocket, which I think are due to the following things:

Point 1: Overall, I think the mapping and the generation of the softcore(sc) potential masks and TI masks works well. There are however sometimes issues with it when it is happening for a growing chain, as in the 42 to 54 ligand transformation and also the 31 to 42 transformation (also table 4,  10.1016/j.ejmech.2013.03.070 ) where the positions of the new H look a bit lost (pictures attached). When doing the reverse, ie 54 to 42, this was not an issue and the run was able to proceed.
42to54
31to42

Additionally, I think that for generating the sc mask a further consideration could maybe be the geometry of the transforming atom. For example, during the  42 to 54 transformation, when the C changes to the N, one of the H that disappears is included in the sc mask as a disappearing atom, and the other H remains the same so is not included in the sc. However, as a result of the C to N transformation and the delocalisation the geometry changes so I think it might be better if the other H were also included in the sc mask for each ligand as well.

Point 2: The amber branch may have broken the normal amber process generation? I noticed this when running the ligand prep; the runs failed with the amber branch but were able to proceed normally with bss-dev. Looking at the input files, this is because the quotation marks around the restraint masks are now mssing, eg
```
#in the amber branch
restraintmask=@0-5,7-19,21-4695,

#in the bss-dev
restraintmask='@1-6,8-20,22-4696',
```
additionally, the indexing is off as it starts at 0 instead of 1. 

Point 3: I think that the amber free energy protocol should also include a minimisation, NVT quilibration and NPT equilibration at each lambda window. This could be an option that could be turned off so that if needed only the files for the actual production run are written, or alternatively be additional steps that need to be included in the workflow. Overall, the system is rarely well enough equilibrated to be able to run without it and otherwise the following error shows up quite alot "ERROR: Calculation halted. Periodic box dimensions have changed too much from their initial values...".  I'll attach the template files I used, although I don't think these necessarily reflect the fastest / best optimised protocol for those preparation steps yet.

Point 4: For most runs, with the current amber.cfg settings the ligand leaves the pocket towards the later windows of the production run. This happened for 54 to 42 and 42 to 55. I checked and alternated through the different settings in the cfg file and also ran longer equilibration simulations before (up to 10 ns) where the ligand remained stable with the Monte Carlo barostat. The default in amber is the Berendsen (barostat = 1), and when this was changed to the Monte Carlo (barostat = 2) in the cfg file the ligand remained in the pocket. Using the MC is in line with what is recommended: (DOI: https://doi.org/10.33011/livecoms.1.1.5957).

Point 5: Input file parameters for the generated amber.cfg
- with previous minimisation and equilibration at each window, ( irest = 1, ntx = 5) are needed in order to restart the simulation using the coordinates and velocities from the previous restart file.
- ntxo is currently written as 1, which is the ASCII format for the final coordinates. The default is (ntxo = 2), which is the NetCDF file format and recommended by the manual as well. Unless there is a particular reason for the ASCII, I think it would be best to have this as the default NetCDF?
- the (logdvdl = 1) needs to be set for a summary of all ∂V/∂λ to be printed at the end of the simulation. I think it would be good to include this as well incase it is needed for post-processing.
- (barostat = 2)  - as in point 4, should be set to this for all production runs.

I'll try and include some of these changes in the branch to then hopefully run a protocol for a complete cycle in the next week?

msuruzhon:
Many thanks for the detailed analysis Anna! Here are my points as well:

> Point 1: Overall, I think the mapping and the generation of the softcore(sc) potential masks and TI masks works well. There are however sometimes issues with it when it is happening for a growing chain, as in the 42 to 54 ligand transformation and also the 31 to 42 transformation (also table 4, [10.1016/j.ejmech.2013.03.070 ](https://www.sciencedirect.com/science/article/pii/S0223523413002304?via%3Dihub)) where the positions of the new H look a bit lost (pictures attached). When doing the reverse, ie 54 to 42, this was not an issue and the run was able to proceed. 42to54 31to42

Isn't this point related to the `merge()` procedure? If so, I guess this impacts all workfloes, not only AMBER. I have also found that the direction of merging matters in terms of defining the new coordinates, but if the mapping is robust, then any problematic bond distances should be taken care of during minimisation. Or is this not what you have observed?

> Additionally, I think that for generating the sc mask a further consideration could maybe be the geometry of the transforming atom. For example, during the 42 to 54 transformation, when the C changes to the N, one of the H that disappears is included in the sc mask as a disappearing atom, and the other H remains the same so is not included in the sc. However, as a result of the C to N transformation and the delocalisation the geometry changes so I think it might be better if the other H were also included in the sc mask for each ligand as well.

This is another good point which is difficult to implement in general (e.g. see [this paper](https://doi.org/10.1021/acs.jctc.0c01328) on how to correctly handle dummy atoms). Especially when there is a change in geometry, it can be difficult to apply single-topology methods and as you said it might be more preferable switching to an approach which perturbs the ring in a dual-topology approach. I am not sure how feasible it would be to automatically detect that within BioSimSpace. Maybe @lohedges and @jmichel80 can weigh in on that?

> Point 2: The amber branch may have broken the normal amber process generation? I noticed this when running the ligand prep; the runs failed with the amber branch but were able to proceed normally with bss-dev. Looking at the input files, this is because the quotation marks around the restraint masks are now mssing, eg (edit: I think the previous push just fixed this!)
> 
> ```
> #in the amber branch
> restraintmask=@0-5,7-19,21-4695,
> 
> #in the bss-dev
> restraintmask='@1-6,8-20,22-4696',
> ```
> 
> additionally, the indexing is off as it starts at 0 instead of 1.

Yes sorry I let that slip for so long. It should be fixed now.

> Point 3: I think that the amber free energy protocol should also include a minimisation, NVT quilibration and NPT equilibration at each lambda window. This could be an option that could be turned off so that if needed only the files for the actual production run are written, or alternatively be additional steps that need to be included in the workflow. Overall, the system is rarely well enough equilibrated to be able to run without it and otherwise the following error shows up quite alot "ERROR: Calculation halted. Periodic box dimensions have changed too much from their initial values...". I'll attach the template files I used, although I don't think these necessarily reflect the fastest / best optimised protocol for those preparation steps yet.

Agreed, I have always used per-lambda protocols so I am in full support of that as well.

> Point 4: For most runs, with the current amber.cfg settings the ligand leaves the pocket towards the later windows of the production run. This happened for 54 to 42 and 42 to 55. I checked and alternated through the different settings in the cfg file and also ran longer equilibration simulations before (up to 10 ns) where the ligand remained stable with the Monte Carlo barostat. The default in amber is the Berendsen (barostat = 1), and when this was changed to the Monte Carlo (barostat = 2) in the cfg file the ligand remained in the pocket. Using the MC is in line with what is recommended: (DOI: https://doi.org/10.33011/livecoms.1.1.5957).

I wasn't aware that Berendsen was the default in AMBER, I thought MC was used in all cases. We will have to change the defaults on that one (maybe Berendsen during equilibration and MC during production to keep it consistent with the other MD engines?). Thanks for that!

> Point 5: Input file parameters for the generated amber.cfg
> 
>     * with previous minimisation and equilibration at each window, ( irest = 1, ntx = 5) are needed in order to restart the simulation using the coordinates and velocities from the previous restart file.

I think this is related to the `FreeEnergy` protocol not supporting restart simulations by default, so the `ConfigFactory` doesn't set these options (otherwise it should). Maybe we can add support for that?

>     * ntxo is currently written as 1, which is the ASCII format for the final coordinates. The default is (ntxo = 2), which is the NetCDF file format and recommended by the manual as well. Unless there is a particular reason for the ASCII, I think it would be best to have this as the default NetCDF?

This is fine by me, I just copied @lohedges protocol. @lohedges is there any particular reason for choosing this as the default?

>     * the (logdvdl = 1) needs to be set for a summary of all ∂V/∂λ to be printed at the end of the simulation. I think it would be good to include this as well incase it is needed for post-processing.

That's another parameter I didn't know existed! I agree, this would be good to add as a default.

>     * (barostat = 2)  - as in point 4, should be set to this for all production runs.

Agreed.
> 
> I'll try and include some of these changes in the branch to then hopefully run a protocol for a complete cycle in the next week?

That sounds great, many thanks for that. I might incorporate some of these changes soon if that's okay, because some of them seem rather critical (e.g. Berendsen), so make sure you are up-to-date with the branch.

lohedges:
Thanks for the detailed insights @annamherz, they are extremely helpful. Just to comment/expand on a few of the points, and @msuruzhon's replies.

Point 1:

Is the non-reversibility issue that you are seeing related to simulations at all lambda windows, and is the problem reproducible over multiple runs? If the system is poorly equilibrated (as you imply it is) then I wouldn't be surprised by this behaviour. If it is reproducible, could you check what happens when you run with SOMD or GROMACS instead. That would rule out something specific to the AMBER FEP protocol, suggesting that something funny has gone on for the merge. However, there are extensive energy tests for the merged system to make sure that the lambda=0/lambda=1 states give the correct energies, etc.

With regards to detecting problematic transformations: this is a tricky approach in general, and is one of the reasons why it's important that users have some understanding of the system that they are attempting to simulate, look at the mappings, etc. I think we are always going to run into some edge cases if people try to automate an entire pipeline over a set of ligands without looking at anything and it's probably futile to try to handle all possible problems. However, within the `merge` function we do have some catches and throw warnings/errors for certain ring related transformations, e.g. size changes, opening/closing, etc. (Here we might suggest doing the transformation in stages.) If there are other problematic merges that can be detected unambiguously, then I'd be happy to flag them and suggest an alternative approach to the user. Obviously we only support single topology at present, so this limits the workarounds that we could propose.

Point 3:

Minimisation/equilibration should be handled with separate protocols, as is the case currently. At present you can pass a perturbable system to a process using _any_ protocl and it will extract and run the lambda = 0 state. From the discussions above, @msuruzhon has implemented FEP specific protocols, e.g. `FreeEnergyEqulibration`, that can run intermediate lambda values. Ultimately I see these being able to be passed to the `FreeEnergy.Relative` constructor to allow you to repeat the protocol for all lambda windows of a given leg. At present this approach hasn't been implemented/tested for GROMACS and SOMD too, so some work needs to be done here before merging. (So there aren't multiple confusing ways of doing the same thing.)

Apologies if I misunderstood, but I thought we would be testing the AMBER FEP implementation as part of the existing workflow used [here](https://github.com/michellab/BioSimSpaceTutorials/tree/main/04_fep), i.e. the one developed by @JenkeScheen and @jmichel80. This uses a different (valid) approach for equilibrating the system, i.e. simulating the individual protein-ligand complexes independently, then extracting the constituents and merging to generate the perturbable system. (I believe a similar approach has been used by Cresset.)

 Point 4:

Thanks for this. I don't have too much expertise with the barostat options in AMBER so I'd be happy to default to whatever is recommended, or the most stable, which is what we've done for the other engines.

Point 5:

I'm happy for restart to be the default option. This should be added as an option to  `BioSimSpace.Protocol.FreeEnergy`, as it already is for `BioSimSpace.Protocol.Production`. Some care will be needed to make sure that it works for SOMD, since it will auto-restart if certain files are present in the working directory. There is already a `_clear_output()` method for `BioSimSpace.Process.Somd`, which could selectively be called if the user passes `restart=False` to the protocol.

Generally I've always written ASCII coordinate files and binary trajectory files. This makes things a bit more portable at the expense of size. However, for AMBER I believe that the parser handles the NetCDF format (see [here](https://github.com/michellab/Sire/blob/devel/corelib/src/libs/SireIO/amberrst.cpp) so we could probably use that directly instead.

Cheers.

JenkeScheen:
> Apologies if I misunderstood, but I thought we would be testing the AMBER FEP implementation as part of the existing workflow used here, i.e. the one developed by @JenkeScheen and @jmichel80. This uses a different (valid) approach for equilibrating the system, i.e. simulating the individual protein-ligand complexes independently, then extracting the constituents and merging to generate the perturbable system. (I believe a similar approach has been used by Cresset.)

I can confirm that Cresset uses the same approach as the one we've implemented (to a degree), although they use OpenMM for this and we use a mixture of sander and pmemd. Also @lohedges, I'd not been made aware that this integration was to be included in the pipeline but that sounds like a great idea, I'd be happy to contribute to that work!

lohedges:
Just another thought for point 1. The non-reversibility could be due to the orignal ligand coordinates, since you will end up with a different result depending on which way you do the alignment, i.e. aligning `ligand0` to `ligand1` based on the mapping, or `ligand1` to `ligand0` based on the inverse mapping. If one confirmation is sub-optimal for some reason that it could potentially result in a crash if there were steric clashes, or the ligand leaving the pocket. You could also see it leave the pocket for one simulation if the ligand that was used as the alignment reference wasn't in the correct binding pose, whereas the reverse is.

annamherz:
> Yes sorry I let that slip for so long. It should be fixed now.

> That sounds great, many thanks for that. I might incorporate some of these changes soon if that's okay, because some of them seem rather critical (e.g. Berendsen), so make sure you are up-to-date with the branch.

Thank you for fixing that and thank you for making those changes! I've checked them all and the output files are generated okay now. I think the restart parameter especially is useful, as it avoids having to set a restart interval.

> Apologies if I misunderstood, but I thought we would be testing the AMBER FEP implementation as part of the existing workflow used [here](https://github.com/michellab/BioSimSpaceTutorials/tree/main/04_fep), i.e. the one developed by @JenkeScheen and @jmichel80.

Yep! I also thought it would eventually be integrated into the existing workflow. I have been following this as an overall framework for the setting up of the FEP, including the ligprep and the equilibration with that before merging and mapping the ligands onto each other in the FEPprep.

> This uses a different (valid) approach for equilibrating the system, i.e. simulating the individual protein-ligand complexes independently, then extracting the constituents and merging to generate the perturbable system.

Still, in the case for amber, running the equilibration with the ligand and protein before is not sufficient eq to be able to run the production runs (failed when trying without additional eq, also is suggested here 10.26434/chemrxiv.14727573.v1 in Figure 7).  What I've been doing for the FEP is carrying out an additional 2 cycle minimisation, NVT and NPT eq at lambda 0.5, before doing a smaller min, NVT, NPT at the individual lambda windows.

When trying to carry this out for the chain growing transformations (42 to 54, 55 to 54, 31 to 42), this prep at 0.5 (at the normal lambda windows if the prep is excluded) is where the issues start. In all cases, after the minimisation, the visualised dual topology ligand and protein look fine, with the H all normal distances away.
However, generally at some point during the NVT or NPT eq, the following error is encountered:
```
Error: an illegal memory access was encountered launching kernel kClearTIKinEnergy_kernel
```
When visualising the final part of the .nc trajectory in VMD before it crashes, they all look fine in terms of distance. Interestingly though, this message shows up in the VMD command line (and doesn't show up for the runs that didn't crash):
```
X_ChangeProperty: BadValue (integer parameter out of range for operation) 0x40
```
Not sure if this is related to the issue as I still need to look at it more closely to see what keeps on causing the crashes, but thought I'd mention it for now!

> If it is reproducible, could you check what happens when you run with SOMD or GROMACS instead. That would rule out something specific to the AMBER FEP protocol, suggesting that something funny has gone on for the merge.

The reverse 54 to 42 and 42 to 31 work, with all other parameters such as the atoms selected for the mask for the ligands etc being kept the same, so the only difference as far as I can tell is the direction of the transformation. Previously, I have been able to run 31 to 42 successfully for somd and gromacs using bss. I checked 42 to 54 with gromacs and somd with the bss setup as well, and these were able to run too. So, I don't think it's necessarily an overall problem with the merging/mapping protocol as a whole or the mask generation and maybe related more to the hybrid dual topology in someway?

> The non-reversibility could be due to the orignal ligand coordinates, since you will end up with a different result depending on which way you do the alignment, i.e. aligning `ligand0` to `ligand1` based on the mapping, or `ligand1` to `ligand0` based on the inverse mapping.

I've been mapping ligand1 to ligand0 as in the execution model. I think though because the issue happens for when 42, 55 and 31 are ligand0, it is unlikely that all of them have a bad conformation, especially when considering that the somd and gromacs are able to run, and that other transformations that don't involve chain growing transformations are still able to run when these are ligand0, for example the 42 to 31 for amber.

Analysis:
The bss analysis script from the execution model failed for amber, and still works for somd and gromacs.
The following error happens:
```
  File "amber_bss_analysis.py", line 59, in <module>
    pmf_bound, overlap_matrix_bound = BSS.FreeEnergy.Relative.analyse(path_to_dir+"/bound")
  File "/home/anna/Documents/clones/BioSimSpace/python/BioSimSpace/FreeEnergy/_relative.py", line 430, in analyse
    raise ValueError("Couldn't find any SOMD, GROMACS or AMBER free-energy output?")
ValueError: Couldn't find any SOMD, GROMACS or AMBER free-energy output?
```
in line 422 of _relative.py, is the following:
```
"AMBER": (Relative._analyse_amber, "/lambda_*/amber.log")
```
The amber output that the _analyse_amber is looking for is actually located in the amber.out . Even when a log file is specified in the command line for the run as follows, it is not written there:
```
$mdrun -i amber.cfg -c heat.rst7 -ref heat.rst7 -p amber.prm7 -O -o amber.out -inf amber.info -e amber.en -r amber.rst7 -x amber.nc -l amber.log
```
However, when changing both cases of amber.log to amber.out in _relative.py I still get the same error of being unable to find the energy output.
I imported it individually to see by making a small change in the BSSanalyseFEP.py from the execution model (around line 58)
```
analyse_amber = Relative._analyse_amber
try:
    pmf_bound, overlap_matrix_bound = analyse_amber(path_to_dir+"/bound")
    pmf_free, overlap_matrix_free = analyse_amber(path_to_dir+"/free")

    freenrg = BSS.FreeEnergy.Relative.difference(pmf_bound, pmf_free)
    freenrg_val = round(freenrg[0].magnitude(), 4)
    freenrg_err = round(freenrg[1].magnitude(), 4)
```
The result is quite different from the one calculated using the alchemlyb implementation (table below).

<div data-pm-slice="1 1 []" data-en-clipboard="true"><u>Final results so far</u></div><div>all in kcal/mol</div>

transformation | experimental | AMBER | GROMACS | SOMD
-- | -- | -- | -- | --
ejm31 to 42 | -0.24 | failed | -0.464 | -0.4051
ejm42 to 31 | +0.24 | +1.998(alchemlyb, BFGS method) -6.8112 (bss amber) | - | -
ejm42 to 54 | -0.76 | failed | -0.6221 | -0.7272


As a final point, the amber branch has broken some stuff for the somd.cfg file generation; the differences between the cfg files being in the following lines:
cfg files generated for 42 to 54:
```
#amber branch
buffered coordinates frequency = 2000.0

#bss-dev
buffered coordinates frequency = 2000
minimise = True
equilibrate = False
```
I fixed the buffered coordinates frequency but when I tried to push the changes it didn't work so I just wanted to check if I have permission to push things to the amber branch?
Otherwise, I think that the minimise=True should probably be kept as this was part of the somd protocol before and seems to work well with it?

lohedges:
The `feature-amber` branch is on a different remote, so you won't have permissions unless Exscientia add you as a collaborator. I agree that it's good to retain the SOMD defaults. Regardless of any pre-miniimisation, we've found that doing a short "minimise to lambda" helps iron out issues with bad contacts, etc.

Thanks again for the detailed comments, it's really helpful. I'll try to digest some of the more technical points and post an update with anything that I think of.

Cheers.

msuruzhon:
Thank you for bringing up these points Anna, here are some thoughts from me:


> Still, in the case for amber, running the equilibration with the ligand and protein before is not sufficient eq to be able to run the production runs (failed when trying without additional eq, also is suggested here 10.26434/chemrxiv.14727573.v1 in Figure 7). What I've been doing for the FEP is carrying out an additional 2 cycle minimisation, NVT and NPT eq at lambda 0.5, before doing a smaller min, NVT, NPT at the individual lambda windows.

I think it makes sense to be running separate lambda-specific equilibrations, since endpoint equilibrations will be biased with respect to the endpoints. It is worrying though that this makes a difference for AMBER but not for the other engines...

> When trying to carry this out for the chain growing transformations (42 to 54, 55 to 54, 31 to 42), this prep at 0.5 (at the normal lambda windows if the prep is excluded) is where the issues start. In all cases, after the minimisation, the visualised dual topology ligand and protein look fine, with the H all normal distances away. However, generally at some point during the NVT or NPT eq, the following error is encountered:
> 
> ```
> Error: an illegal memory access was encountered launching kernel kClearTIKinEnergy_kernel
> ```
> 
> When visualising the final part of the .nc trajectory in VMD before it crashes, they all look fine in terms of distance. Interestingly though, this message shows up in the VMD command line (and doesn't show up for the runs that didn't crash):
> 
> ```
> X_ChangeProperty: BadValue (integer parameter out of range for operation) 0x40
> ```
> 
> Not sure if this is related to the issue as I still need to look at it more closely to see what keeps on causing the crashes, but thought I'd mention it for now!

Could this be related to a badly compiled AMBER executable? Also, which pmemd version are you using? I also just realised that setting `logdvdl=1` results in errors during minimisation for me, so I turned it off now by default (you don't need that information outside of production anyway). Does that fix anything or...?

> The reverse 54 to 42 and 42 to 31 work, with all other parameters such as the atoms selected for the mask for the ligands etc being kept the same, so the only difference as far as I can tell is the direction of the transformation. Previously, I have been able to run 31 to 42 successfully for somd and gromacs using bss. I checked 42 to 54 with gromacs and somd with the bss setup as well, and these were able to run too. So, I don't think it's necessarily an overall problem with the merging/mapping protocol as a whole or the mask generation and maybe related more to the hybrid dual topology in someway?

That's also very worrying, we should make sure that AMBER and GROMACS/SOMD result in the same free energies using the same force field.

> Analysis: The bss analysis script from the execution model failed for amber, and still works for somd and gromacs. The following error happens:
> 
> ```
> Traceback (most recent call last):
>   File "amber_bss_analysis.py", line 59, in <module>
>     pmf_bound, overlap_matrix_bound = BSS.FreeEnergy.Relative.analyse(path_to_dir+"/bound")
>   File "/home/anna/Documents/clones/BioSimSpace/python/BioSimSpace/FreeEnergy/_relative.py", line 430, in analyse
>     raise ValueError("Couldn't find any SOMD, GROMACS or AMBER free-energy output?")
> ValueError: Couldn't find any SOMD, GROMACS or AMBER free-energy output?
> ```
> 
> in line 422 of _relative.py, is the following:
> 
> ```
> "AMBER": (Relative._analyse_amber, "/lambda_*/amber.log")
> ```
> 
> The amber output that the _analyse_amber is looking for is actually located in the amber.out . Even when a log file is specified in the command line for the run as follows, it is not written there:
> 
> ```
> $mdrun -i amber.cfg -c heat.rst7 -ref heat.rst7 -p amber.prm7 -O -o amber.out -inf amber.info -e amber.en -r amber.rst7 -x amber.nc -l amber.log
> ```
> 
> However, when changing both cases of amber.log to amber.out in _relative.py I still get the same error of being unable to find the energy output. I imported it individually to see by making a small change in the BSSanalyseFEP.py from the execution model (around line 58)
> 
> ```
> analyse_amber = Relative._analyse_amber
> try:
>     pmf_bound, overlap_matrix_bound = analyse_amber(path_to_dir+"/bound")
>     pmf_free, overlap_matrix_free = analyse_amber(path_to_dir+"/free")
> 
>     freenrg = BSS.FreeEnergy.Relative.difference(pmf_bound, pmf_free)
>     freenrg_val = round(freenrg[0].magnitude(), 4)
>     freenrg_err = round(freenrg[1].magnitude(), 4)
> ```
> 
> The result is quite different from the one calculated using the alchemlyb implementation (table below).
> Final results so far
> all in kcal/mol
> transformation 	experimental 	AMBER 	GROMACS 	SOMD
> ejm31 to 42 	-0.24 	failed 	-0.464 	-0.4051
> ejm42 to 31 	+0.24 	+1.998(alchemlyb, BFGS method) -6.8112 (bss amber) 	- 	-
> ejm42 to 54 	-0.76 	failed 	-0.6221 	-0.7272

We probably have different versions of AMBER, since mine outputs everything in the `amber.log` file. Could you perhaps share your output file? In any case, it looks like I should rewrite the `analyse()` function to depend on `alchemlyb` and add one extra dependency? I did test my MBAR calculation against `alchemlyb` but maybe something else has gone wrong.

> As a final point, the amber branch has broken some stuff for the somd.cfg file generation; the differences between the cfg files being in the following lines: cfg files generated for 42 to 54:
> 
> ```
> #amber branch
> buffered coordinates frequency = 2000.0
> 
> #bss-dev
> buffered coordinates frequency = 2000
> minimise = True
> equilibrate = False
> ```

That should be fixed now.

> I fixed the buffered coordinates frequency but when I tried to push the changes it didn't work so I just wanted to check if I have permission to push things to the amber branch? Otherwise, I think that the minimise=True should probably be kept as this was part of the somd protocol before and seems to work well with it?

It might make sense to open small pull requests on our branch in the future to do so, as I don't think it will be straightforward to grant you access on our branch. 

`minimise=True` is now only turned on as part of the `FreeEnergyMinimisation` protocol, to keep it consistent with the other MD engines. Would there be any differences to using a preliminary `FreeEnergyMinimisation`  instead?

Many thanks.

annamherz:
> I think it makes sense to be running separate lambda-specific equilibrations, since endpoint equilibrations will be biased with respect to the endpoints. It is worrying though that this makes a difference for AMBER but not for the other engines...

For gromacs I've also been using a min, NVT and NPT at each lambda window as well. I've tried running it without (for both 31 to 42 and 42 to 54) , and got errors because the system was unstable for the complex for 31 to 42, and for 42 to 54 the run is excruciatingly slow at about 12 ns/day whereas with the eq steps before it is about 70 ns/day.
The main reason I've been doing the 0.5 lambda window extra eq for amber was to see if that could fix the kernel error I was getting, but unfortunately that didn't help.

> Currently there is no way to decouple the minimisation / equilibration from the production run within SOMD and keep all of the other functionality, e.g. logging, consistent. The minimisation and equilibration routines (those within `somd-freenrgy`that run an intermediate lambda state) are _unique_ to SOMD, provide no logging of thermodynamic information, and don't report any trajectory or restart information, i.e. they are just an optional pre-step of the main production FEP run. In order for them to work correctly with your mixin approach I'll need to work out how to isolate them from the SOMD code, and also adapt them so they actually report the expected information like the other MD engines do. At present you might be able to set `minimise=True`, or `equilibrate=True`, then run a one step / single cycle FEP simulation at a given lambda value. However, you won't get any useful logging info, or a trajectory, so things like `report_interval` and `restart_interval` would be meaningless for this approach.

Oh I didn't know that! I think because of this then it's probably easiest to leave the somd free energy protocol with minimise automatically set to true for now? 

> This is interesting. It might be that there isn't a one size fits approach to minimisation and equilibration for FEP simulations with different engines. If so, we might need to abstract things a bit further and decouple the FreeEnergyMinimisation and FreeEnergyEquilibration protocols from their non-FEP equivalents, i.e. we just end up doing what is most robust / efficient for the underlying engine. This could also mean that you will potentially need to do the _entire_ pipeline, i.e. FEP minimisation, equilibration, production, with the same engine. At present, with the lambda=0 or BioSimSpaceTutorials approach, we have taken advantage of using faster engines for minimisation and equilibration, e.g. running with `pmemd`. That said, we have run into several issues where an "equilbrated" system with one engine immediately blows up when used as the input for a production run with another. This is one reason why we always leave the `minimise=True` option on within SOMD, since OpenMM seems to be particularly sensitive to the starting confirmation. (Perhaps other engines automatically adjust bad contacts on start up, or have different tolerances.)

I think overall having the ligprep is still a good idea because it makes sure that all the different engines start from a reasonable structure in terms of merging and takes care of the MD part with regards to the ligand being okay in the binding pocket just by itself.
Then afterwards, for somd having just the free energy protocol with the minimisation set to true as it seems to be able to handle everything okay with just this.
For amber and gromacs, I think they will have to have a seperate series of protocols such as the min, NVT, NPT and then free energy production run (as there is no way to in-cooperate this into a single input file). I think it would probably be most robust if these are run at each lambda window with their own respective engine to avoid issues with the simulation blowing up and/or crashing.
I guess in terms of the execution model, this would mean that during the FEPprep script the merging at the start stays the same and at the end a different series of protocols are decided upon depending on which engine was selected?

>We probably have different versions of AMBER, since mine outputs everything in the amber.log file. Could you perhaps share your output file? In any case, it looks like I should rewrite the analyse() function to depend on alchemlyb and add one extra dependency? I did test my MBAR calculation against alchemlyb but maybe something else has gone wrong.

I've been using amber20? I've attached one of the output files from the 42 to 31 run. I've just realised that there is a mistake and it's from before the restart was specified for the final run, so I'll rerun the production windows for that transformation with the restart and maybe that will also change the calculated MBAR value.
[amber.zip](https://github.com/michellab/BioSimSpace/files/7602987/amber.zip)

> Could this be related to a badly compiled AMBER executable? Also, which pmemd version are you using? I also just realised that setting logdvdl=1 results in errors during minimisation for me, so I turned it off now by default (you don't need that information outside of production anyway). Does that fix anything or...?

yes logdvdl should be turned off for everything but the production run, so I think it should only be set that it is on if its the FreeEnergy.Relative(). As I've still been using the templates I had for the min and eq, that shouldn't be what's causing the issues for me. The pmemd version I'm using is the GPU (CUDA) Version 18.0.0  with a nvidia GPU. I'll try recompiling amber to see if that's what is causing the problems, and once I've done that try running the previous runs again.

> That should be fixed now.

great thank you!

annamherz:
Hello! I have finished running all the transformations once. I have also recompiled amber, but the runs that failed before still failed so I don't think that was the issue.

#### initial results
The protocol is based on that in the execution model, and the same starting files from ligprep are used for all runs. 
SOMD as run as in execution model, with just the minimise=True.
GROMACS with additional min (5000 steps), NVT (100ps), NPT (100ps) at each lambda window, generated using itemplates outside of BSS. Production input files were prepared using BSS.
For AMBER, intially 2 step minimisation (first restraining the backbones, then nothing), NVT (1000ps with heating), NPT (1000ps) were carried out at lambda=0.5 and this equilibrated state was used as the starting point for the lambda windows. As mentioned before, this was mainly done initially to see if it could stop the runs from failing (Error: an illegal memory access was encountered launching kernel kClearTIKinEnergy_kernel), however equilibration seemed to have not been the issue. When the sc mask also contained the transforming atoms (as opposed to just the dummy, as currently generated by the branch), these runs did not fail. At each lambda window, an additional  min (5000 steps), NVT (50ps), NPT (50ps), generated using templates outside of BSS, were carried out also. Production input files were prepared using BSS.
Analysis was carried out using the analysis script from the BSS execution model and the error is that estimated by pymbar (other than one of the gromacs runs). Each run was carried out once.
The results of all the initial runs are attached in the table:
[initial_runs.xlsx](https://github.com/michellab/BioSimSpace/files/7641923/initial_runs.xlsx)

Based on the results here I think the default should be to also include the transforming atoms in the scmask as well as this seems less likely to fail, there is a smaller error in the mbar and overall the results seem closer to the experimental. Also, the following paper (J. Chem. Inf. Model. 2020, 60, 11, 5296–5300) similarly compared two protocols, one with defining sc region using the maximum common substructure and then also as grouping them into chemical functional groups. Using the second was suggested to be slightly better as "there are no softcore transformations involving H atoms that do not also consider at the same time the heavy atoms to which they are covalently attached".  In their case of just including the dummy atoms in the sc there were numerical integration errors for the methanol to ethane transformation. This was avoided if the whole CH3 and OH are considered as the sc regions respectively, and this approach was also found to have higher statistical precision. I think overall having the dummy and transforming atoms in the sc mask then would be more robust as a default protocol.

jmichel80:
Pasting Anna's results here to facilitate discussion 

image

@annamherz  - can you give details on how the 'error' is calculated for each method ? Is it the same approach or does it vary between code ? What are the settings used by pymbar when run from within BSS to perform error estimation ?  The Amber 'error' estimates are surprisingly large. Also ejm31-->ejm42 with SOMD. The SOMD error estimate doesn't seem reliable for that one as your free energy change is actually fairly close to the value reported in the 'other run' column. 

Also clarify what code was used for the 'other runs' figures you report. 

Could you also post the somd.cfg , equivalent Gromacs config settings and AMBER config settings for the actual simulations.  In particular I would like to clarify which integration time step, HMR and constraint settings you use with each code. We expect it should be feasible to use by default a 4 fs integration time step at least for SOMD and AMBER.  

@msuruzhon  it would be useful to know what free energy estimates you obtain on your end for the above transformations.

msuruzhon:
Thank you @annamherz and @jmichel80 for this useful benchmark. Right now the results seem worryingly off for AMBER and this is something that we should investigate - maybe there are some default parameters that I haven't set correctly. I will try to reproduce these results as well and will get back to you once I have some data.

annamherz:
> Also clarify what code was used for the 'other runs' figures you report.

for the other runs the following was used:
(b) is from the BSS execution model on github, which was run in SOMD using the default parameters for free energy, 7 windows, each at 4ns.
(c) is run using AMBER20 and averaged over 10 runs, using SSC(2) (smoothstep softcore) and GAFF2. I couldn't really find any information of how many windows / length of run / etc. exactly, but it says it used the protocols as described in the best practices guide, so I'm assuming with MC barostat, langevin temperature dynamics.

> Could you also post the somd.cfg , equivalent Gromacs config settings and AMBER config settings for the actual simulations. In particular I would like to clarify which integration time step, HMR and constraint settings you use with each code. We expect it should be feasible to use by default a 4 fs integration time step at least for SOMD and AMBER.

Attached are the input files I used. For the template files for gromacs, %W% was substituted with the relevant lambda window and for the amber files %L%, as well as the %MX% substituted with the masks from the production run. The production input files are written using bss except where I modified the scmasks to include the transforming atoms as well. Overall 2fs and no HMR restraints were used as this is the default that is currently generated.
[input_files.zip](https://github.com/michellab/BioSimSpace/files/7660775/input_files.zip)

> can you give details on how the 'error' is calculated for each method ? Is it the same approach or does it vary between code ? What are the settings used by pymbar when run from within BSS to perform error estimation ? The Amber 'error' estimates are surprisingly large

Looking at the BSS code, analyse_gromacs makes use of gmx bar, analyse_somd uses somd-freenrg which uses pymbar and analyse_amber uses pymbar as well. The error calculated is from propagation of the errors of the free and bound transformation from the last window in pymbar.

I ran repeats for amber and have the following results:

transformation | RBFE (kcal/mol) | error (kcal/mol) | average (incl first run) (kcal/mol) | error (SEM) | experimental (kcal/mol)
-- | -- | -- | -- | -- | --
ejm31 to 42 run 2 | 0.5690 | 0.5969 | +0.3549 | 0.1139 | -0.24
ejm31 to 42 run 3 | 0.1804 | 0.6243
ejm42 to 31 run 2 | -0.0624 | 0.5403 | +0.181 | 0.295 | +0.24
ejm42 to 31 run 3 | -0.1627 | 0.5590
ejm42 to 54 run 2 | -0.4145 | 0.6343 | -0.3466 | 0.1895 | -0.76
ejm42 to 54 run 3 | -0.6356 | 0.5910
ejm42 to 55 run 2 | +0.8991 | 0.4649 | +0.624 | 0.225 | +0.57
ejm42 to 55 run 3 | +0.1784 | 0.4234
ejm54 to 42 run 2 | +0.9015 | 0.6787 | +1.248 | 0.204 | +0.76
ejm54 to 42 run 3 | +1.6068 | 0.6729
ejm55 to 54 run 2 | -1.1084 | 0.6928 | -0.783 | 0.323 | -1.32
ejm55 to 54 run 3 | -0.1363 | 0.7326
cycle closure 42-55-54 | - | - | 1.09 | - | 0.01

The individual errors from pymbar are similar as before. When I calculated some of the runs outside of bss, using the alchemlyb library, there were similarly large errors so I don't think it is a problem with the way bss analyses the data.
The alchemlyb library is able to analyse the gromacs output using pymbar, so I think it may be better to use pymbar for the gromacs runs as well instead of the gmx bar so the analysis methods between the different implementations are more consistent. Ideally, if it's possible to write all output for the different engines into the same format (energy matrix and no of samples) and then have that processed by pymbar that would probably give more consistency in the analysis?

lohedges:
Thanks, @annamherz.

> Ideally, if it's possible to write all output for the different engines into the same format (energy matrix and no of samples) and then have that processed by pymbar that would probably give more consistency in the analysis?

Yes, I think this would be a good idea in the long term. Originally we had thought that BioSImSpace might be a bit more modular, i.e. you wouldn't need all the components to do something and it would just use what it found. As such, we decided to use the _standard_ analysis tool for the given FEP engine. In practice, it turns out to be a bit of a maintenance nightmare to do things this way so we've ended up standardising things and bundling software (via conda dependencies) where possible (other than the actual MD engines, which need to be built for specific hardware for good optimisation). If it's possible to do everything via `alchemlyb` or `pymbar`, then it would certainly be better to do it that way.

@msuruzhon: Apologies that I've not had time to do a formal review of the code yet. In light of the AMBER discrepancies I think it's probably best that we focus on figuring that out first. I might try creating a copy of your branch in our repo and merging things from devel into that so that I can begin to resolve some of the conflicts.

msuruzhon:
Hi @jmichel80 @annamherz @lohedges, I ran the perturbations used in the AMBER tutorial using the input files the York group uses, and these are the results I get with SOMD:

edge | ddG | error | Anna | Lee *et al.*
-- | -- | -- | -- | --
ejm42~ejm54 | -1.63 | 0.11 | -0.73 | -1.76 
ejm54~ejm42 | 1.48 | 0.11 |  0.32 |
ejm42~ejm55 | -0.92 | 0.06 | 0.17 | -1.18
ejm55~ejm42 | 0.97 | 0.06 | | |
ejm54~ejm55 | 0.99 | 0.10 | | | 
ejm55~ejm54 | -0.68 | 0.10 |  0.02 |-0.53

You can see that these energies are similar to the results reported by [Lee *et al.*](https://pubs.acs.org/doi/abs/10.1021/acs.jcim.0c00613) but they seem to be very different to the ones obtained by @annamherz . Could this difference be attributed to the different input files we have used? I can send around some serialised `BioSimSpace` objects that I have used for the simulations if it helps.

jmichel80:
Hi @msuruzhon can you share details on how you converted the York input files into SOMD inputs ? As you said there may well be a difference in the inputs that explains this result. Did you check whether you can reproduce Anna's results using her inputs ? That would be help ensure that we are running the same software stacks.

msuruzhon:
Hi @jmichel80, the input files I sent around some time ago are completely prepared and solvated and the only thing that needs to be done is to merge the two ligands in the `BioSimSpace` object. The input files I used are based on the files I sent but they have been solvated and pre-equilibrated using our internal workflows, so I am attaching the serialised objects I have used to start the simulations in `BioSimSpace`. In all cases I am running 11 lambda windows at 5 ns each with a short per-lambda 20-step minimisation and unrestrained 50 ps equilibration at 4 fs with HMR.

I can't find Anna's input files, maybe you could share them so that I could also try to investigate why we observe these differences?

Many thanks.

[Archive Part 1.zip](https://github.com/michellab/BioSimSpace/files/7814511/Archive.Part.1.zip)
[Archive Part 2.zip](https://github.com/michellab/BioSimSpace/files/7814524/Archive.Part.2.zip)

lohedges:
Also check that you are both using the same version of SOMD, i.e. Sire and BioSimSpace. There have been a few fixes lately that _could_ affect results depending on the terms in the perturbation.

jmichel80:
Hi @msuruzhon , thanks for sharing the inputs. @annamherz if you could take a look that would be helpful, please also share your SOMD inputs. 
@msuruzhon can you share the exact somd.cfg file used for minimisation, equilibration, production ? We haven't run with 4 fs internally yet, in our work with Cresset we do but this uses a Langevin Integrator that isn't exposed (yet) in our Sire devel branch.

msuruzhon:
Thanks @jmichel80 and @lohedges, I am attaching the configuration files I used for the runs. The Sire version I use is `sire==2021.1.0=py37hf484d3e_1`. The previously attached serialised objects already have modified masses so they are ready to run at 4 fs.

Thanks.

[Configuration files.zip](https://github.com/michellab/BioSimSpace/files/7814684/Archive.zip)

annamherz:
Hello! Thank you for the results!

> Hi @jmichel80, the input files I sent around some time ago are completely prepared and solvated

> I can't find Anna's input files, maybe you could share them so that I could also try to investigate why we observe these differences?

At the start when I was testing for amber, the ligand kept on leaving the pocket (which ended up being unrelated to the ligand inputs but due to the barostat) but at some point I started from scratch and used the 4GIH pdb entry and edited/docked the other ligands in flare before following the start of the BSS tutorial for ligand prep, which included a 2000 step minimisation 50ps NVT and 200ps NPT equilibration for each the complex / solvated ligand. I've been using those input files since then for the merging, etc.:

[prep.zip](https://github.com/michellab/BioSimSpace/files/7821229/prep.zip)

> Also check that you are both using the same version of SOMD, i.e. Sire and BioSimSpace

> The Sire version I use is sire==2021.1.0=py37hf484d3e_1.

the version of Sire I've been using is 2021.1.0 py37hf484d3e_21 .

> I am attaching the serialised objects I have used to start the simulations in BioSimSpace. In all cases I am running 11 lambda windows at 5 ns each with a short per-lambda 20-step minimisation and unrestrained 50 ps equilibration at 4 fs with HMR.

Great thank you! I'll have a look at those and the cfg files. I was also running 11 windows, but shorter at 2ns with a 2fs timestep without HMR.

msuruzhon:
Thanks @annamherz, I suspect most of the differences will be due to the different starting configurations.

I finally managed to get some satisfactory results with AMBER. One of the problems with it was the fact that 4 fs simulations (HMR) or 2 fs simulations (no HMR) crashed when SHAKE was turned off for the ligand. This is obviously not ideal, and the solution I have now is that I use a pruned MCS which removes all hydrogen atoms which are perturbed to non-hydrogen atoms, effectively resulting in two extra soft-core atoms per perturbed constraint. @annamherz, I think this is consistent with your observations. This does mean, however, that we should probably change the `hybrid` mapping to a mapping that prunes the `single` mapping, instead of only mapping atoms to the same element (the current implementation), which seems to be overly restrictive. @annamherz, @lohedges and @jmichel80, any thoughts on that?

The second problem was related to some of the default parameters. One of them is the PBC wrapping parameter `iwrap` which should be set to 1, otherwise the coordinates can result in numerical overflows and AMBER (or even BioSimSpace during I/O) can crash without much detail. Another observation I made is that the energy output interval parameter `ntpr` should be made equal to e.g. 100 steps, instead of `report_interval` steps, otherwise a very low number of energies are output. Multiples of 50 such as 250 steps didn't work very well for me for some reason (some of the output steps were skipped for no apparent reason), so I think we need to stick to multiples of 100 steps. Since the default for SOMD and GROMACS is 250 steps, we should think about how to make these defaults consistent with AMBER. @jmichel80 and @lohedges, I would appreciate your feedback on that.

The third problem I had was related to me setting `iwrap` to 1. In some cases this results in one of the ligands having different periodic image coordinates to the other ligand. Since I assume that both of them have the same periodic image in the `getSystem()` function, this resulted in some very distorted structures and a fairly tricky to pin down bug. However, this should be fixed now.

With all this in mind, I managed to run the 6 perturbations with 4 fs timestep from the same initial structures as the SOMD run, and these are the results I get:

edge | ddG SOMD | error SOMD | ddG AMBER| error AMBER | \|SOMD - AMBER\| | Anna SOMD | Lee et al. AMBER
-- | -- | -- | -- | -- | -- | -- | --
ejm42~ejm54 | -1.63 | 0.11 | -1.71 | 0.03 | 0.08 | -0.73 | -1.76
ejm54~ejm42 | 1.48 | 0.11 | 1.69 | 0.03 | 0.21 | 0.32 |  
ejm42~ejm55 | -0.92 | 0.06 | -1.22 | 0.02 | 0.30 | 0.17 | -1.18
ejm55~ejm42 | 0.97 | 0.06 | 1.29 | 0.02 | 0.32 | |
ejm54~ejm55 | 0.99 | 0.10 | 0.83 | 0.02 | 0.15 | |  
ejm55~ejm54 | -0.68 | 0.10 | -0.70 | 0.02 | 0.03 | 0.02 | -0.53

The maximum deviation between SOMD and AMBER is 0.30 kcal/mol, which I would say is pretty good - maybe @jmichel80 will also have some insights on what sort of reproducibility is satisfactory?

Many thanks.

lohedges:
> This does mean, however, that we should probably change the hybrid mapping to a mapping that prunes the single mapping, instead of only mapping atoms to the same element (the current implementation), which seems to be overly restrictive. @annamherz, @lohedges and @jmichel80, any thoughts on that?

I'll have a think and get back to you.

> One of them is the PBC wrapping parameter iwrap which should be set to 1, otherwise the coordinates can result in numerical overflows and AMBER (or even BioSimSpace during I/O) can crash without much detail.

I'm happy with having `iwrap=1` as the default. The only reason that it's not is that it's trivial to wrap the coordinates after the fact, but non-trivial (or impossible) to recover the _actual_ imaged coordinates if the trajectory spacing is sufficiently large. In previous work I was interested in dynamics of slow systems, so needed to measure displacements over logarithmically increasing time scales. I've never understood why coordinates aren't wrapped to the minimum image, with three integers used to indicate the _actual_ image in the x, y, and z dimensions. (These could be output alongside the minimum image coordinates so you can reconstruct displacements.) I'm not sure how wrapping is handled for all of the other engines (or what the default is) so it would be good if we were consistent here.

> Another observation I made is that the energy output interval parameter ntpr should be made equal to e.g. 100 steps, instead of report_interval steps, otherwise a very low number of energies are output

This is fine too. I chose 250 to match the well tested Cresset implementation, but it shouldn't be too sensitive, other than the restart interval needing to a multiple of it. (This would need to be adjusted in the SOMD config generator.)

Cheers.

jmichel80:
> Another observation I made is that the energy output interval parameter ntpr should be made equal to e.g. 100 steps, instead of report_interval steps, otherwise a very low number of energies are output. Multiples of 50 such as 250 steps didn't work very well for me for some reason (some of the output steps were skipped for no apparent reason), so I think we need to stick to multiples of 100 steps. Since the default for SOMD and GROMACS is 250 steps, we should think about how to make these defaults consistent with AMBER. @jmichel80 and @lohedges, I would appreciate your feedback on that.

It shouldn't be a problem to set this parameter by default to 200.

>The maximum deviation between SOMD and AMBER is 0.32 kcal/mol, which I would say is pretty good for a binding FE calculation - maybe @jmichel80 will also have some insights on what sort of reproducibility is satisfactory?

That seems reasonable. 
It would be good to check that you get reliable and consistent results by starting parameterisation/solvation/equilibration from scratch (from a sensibly chosen set of starting structures) to make sure that we haven't overlooked something else in the pipeline.

msuruzhon:
Hi @jmichel80 @annamherz I uploaded the pre-equilibrated Tyk2 input files that I am going to be using for the benchmark in [this repo](https://github.com/Exscientia/RBFEBenchmark). Please let me know if there are any questions.

annamherz:
Hello! I've merged the changes into a new amber branch in the michellab/BioSimSpace to bring it more up-to-date with the current development branch and resolve some of the conflicts there, and also so I can commit changes more easily when testing.

lohedges:
Thanks, @annamherz, this will be really useful. Note that I've renamed the branch to `feature-amber-fep` so that it matches the repository naming conventions. To update your local branch you can run (assuming your local branch isn't ahead):

```
git fetch
git checkout feature-amber-fep
git branch -D amber
```

When I get a chance I'll go through to make sure that anything hasn't been missed during the merge. Since things have diverged a regular merge and conflict resolution might be difficult. Ideally we need to accept everything from `devel`, then apply the changes from the remote in the context of the updates to devel. (For example, there are several places where the new `try_import` functionality should be applied to deal with dependencies that aren't yet available on certain platforms.)

On importing your branch I see the following errors:

```python
~/Code/BioSimSpace/python/BioSimSpace/FreeEnergy/_relative.py in <module>
     45 import numpy as _np
     46 from simtk import unit as _unit
---> 47 from simtk.unit.constants import MOLAR_GAS_CONSTANT_R as _MOLAR_GAS_CONSTANT_R
     48 import pymbar as _pymbar
     49

ModuleNotFoundError: No module named 'simtk.unit.constants'
```

This is because of the update to the conda-forge OpenMM, where the import structure has changed slightly. (You now import OpenMM directly, rather than using `simtk.openmm`, for example.) However, we have our own units within Sire so (to avoid depending on an external package) it could be changed to.

```python
from Sire.Units import gasr as _gasr
```

Perhaps @msuruzhon could try merging your branch back into his (or another branch) to see if things still work for him. This would give him access to the most recent changes on devel too.

With regards to ultimately merging all of this over to devel: I intend to go through the code and merge things over in chunks with separate pull requests. Firstly I'd like to port over the new config generation code, making sure that it is isolated in its own sub-package and is as interoperable as possible. This will also mean adding config generation support for NAMD too.

Thanks again for taking the time to merge. I hope it wasn't too much of a pain and let me know if you have any questions or concerns about the result, since I have a good handle on what has been changed in devel of late, and what _should_ be correct.

Cheers!

annamherz:
> To update your local branch you can run (assuming your local branch isn't ahead):

Great, thank you! 

> This is because of the update to the conda-forge OpenMM, where the import structure has changed slightly. (You now import OpenMM directly, rather than using `simtk.openmm`, for example.) However, we have our own units within Sire so (to avoid depending on an external package) it could be changed to.

Okay, I was unsure about the best way to go about this. I'll make the changes with the Sire units and commit them later today / tomorrow.

> Thanks again for taking the time to merge. I hope it wasn't too much of a pain and let me know if you have any questions or concerns about the result, since I have a good handle on what has been changed in devel of late, and what _should_ be correct.

Thank you!

msuruzhon:
Many thanks for the work @annamherz and @lohedges ! I will test the new branch next week and will report back if anything has stopped working as expected. Cheers.

PR #245, opened 2021-11-04
Lomap improvements again

JenkeScheen:
some more tweaks to `generateNetwork()` that were necessary for my work. 

1) `molecules` can now also be passed as a list of RDKit molecules instead of BSS molecules. In this case, the intermediate ligand files written to `work_dir` are SDF. The reason for this is that for ~60% of test cases (FEP congeneric series), RDKit kekulization fails when using MOL2 inputs, i.e. making LOMAP fail. See some other discussions: https://github.com/MobleyLab/Lomap/issues/26 and https://sourceforge.net/p/rdkit/mailman/message/35999854/
Using an SDF intermediate makes all molecules pass; presumably this is what Cresset is doing as well (see Max's comment in the linked LOMAP issue).
2) I'm generating a large number of networks for many different series and I want to compare networks after letting my code run for a few days, so I have a use-case where I want to write network plots outside of a notebook. This PR changes `generateNetwork()` such that it writes the plot file to `work_dir` even when not in a notebook - if in a notebook, the plot it still shown.

I realise that these types of PRs are not very sustainable wrt ongoing work over at Mobley et al; perhaps all my PRs here are making our implementation too convoluted? I'm not sure who else even uses our current LOMAP implementation.. Anyway, this is just for posterity, if you'd rather I fork my own I'd be happy to do that too instead.

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -155,7 +155,6 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
     # A list of RDKit molecule objects.
     elif all(isinstance(x, _Chem.rdchem.Mol) for x in molecules):
         rdkit_input = True
-
     else:
         raise TypeError("'molecules' must be a list of "
                         "'BioSimSpace._SireWrappers.Molecule' "
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -235,17 +235,15 @@ def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
     _os.makedirs(work_dir + "/outputs", exist_ok=True)
 
     # Write all of the molecules to disk.
-    if not rdkit_input:
-      for x, (molecule, name) in enumerate(zip(molecules, names)):
-          _IO.saveMolecules(work_dir + f"/inputs/{x:03d}_{name}",
-              molecule, "mol2", property_map=property_map)
-
-    elif rdkit_input:
-      for x, (molecule, name) in enumerate(zip(molecules, names)):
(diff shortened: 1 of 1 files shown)


lohedges:
Thanks, I'll be able to look this next week. I'm happy for you to keep tweaking and improving this functionality as I'm sure it'll be better for other users in the long run. Don't worry too much about the Lomap divergence, there's not too much we can do until they refactor things at their end and make a conda package.

Cheers.

lohedges:
Thanks, I'll be able to look this next week. I'm happy for you to keep tweaking and improving this functionality as I'm sure it'll be better for other users in the long run. Don't worry too much about the Lomap divergence, there's not too much we can do until they refactor things at their end and make a conda package.

Cheers.

lohedges:
Hi @JenkeScheen. This looks good. I hope you don't mind but I jumped on the branch to make a few formatting tweaks. Mostly to make the indentation consistent (4 spaces) and refactor a few conditional blocks for readability. Let me know if it still works at your end and I'll proceed with the merge.

Cheers.

JenkeScheen:
Hi @lohedges, these look freat. Yes, sorry about the indentation - I think Sublime Text has had a stealthy update, I've been having issues with spaces vs tabs for weeks now. Thanks for tweaking the PR!

lohedges:
No problem. While Python gets very angry about whitespace and indentation, it actually allows you to mix indentation levels as long as the indentation is consistent within each unique block. This is annoying when you switch developers / editors since you can break the code when making edits by incorrectly indenting an existing block. (And since it's ot compiled, the bugs can be subtle runtime issues.) This is why I prefer everything being consistent from the outset.

lohedges:
(And is why I much prefer curly braces, since the scope is made explicit, and you can format things however you like within that scope.)

PR #255, opened 2021-12-08
Feature lomap force n edges

JenkeScheen:
this PR adds functionality to `BSS.Align._align.generateNetwork()` such that it now ingests a user-specified integer (`n_edges_forced`) that forces the number of edges included in the perturbation network. This is useful for general practice as typically FEP campaigns have a pre-set amount of compute available. It also allows direct comparison between networks generated with different types of inputs (i.e. beside LOMAP-Score using the `links_file`).

I have included validation (checks that `n_edges_forced` is an _int_ and checks that it is higher than 0 and at most a fully-connected network.

There are three situations as described in the code:
1. the user-specified number of edges is already suggested by LOMAP, in this case do nothing.
2. the user-specified number of edges is higher than in the LOMAP network; in this case add n top-scoring edges parsed from the LOMAP output file.
3.  the user-specified number of edges is lower than in the LOMAP network; in this case remove n lowest scoring edges from the suggested network. This situation is discouraged as it can cause network cycles to break and also ligands/clusters to become disconnected from the network. Although it is possible to attempt to remove edges that 'do the least damage' (e.g. prioritising cycle edges), this would have to be done directly in the LOMAP API and could be done in later work. It's still a useful feature IMO as it allows users to run FEP on the most reliable sub-network(s) of the series.

For all cases the network is still plotted correctly, although in 3) there can of course be disconnected parts of the network drawn. In 3), completely disconnected nodes are excluded from the network drawing.

### DIFF
--- a/python/BioSimSpace/Align/_align.py
+++ b/python/BioSimSpace/Align/_align.py
@@ -80,7 +80,7 @@ except:
     _fkcombu_exe = None
 
 def generateNetwork(molecules, names=None, work_dir=None, plot_network=False,
-        links_file=None, property_map={}):
+        links_file=None, property_map={}, n_edges_forced=None):
     """Generate a perturbation network using Lead Optimisation Mappper (LOMAP).
 
        Parameters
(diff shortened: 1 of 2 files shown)


lohedges:
Thanks, @JenkeScheen. This sounds really useful. I should be able to take a look at some point this afternoon and let you know if I have an questions / comments before merging.

JenkeScheen:
thanks @lohedges. Thinking about this some more, the comment I had on 3) kind of also counts for 2), i.e. in its current form the 'top-scoring' non-included edges are included for this scenario - in a more involved algorithm we would probably want a clever selection of which edges to add, e.g. by reinforcing troublesome nodes. Again, this would have to be done in the LOMAP API directly and is out of scope for this PR.

lohedges:
All looks good. I've just made a few formatting tweaks and trimmed some whitespace.

Cheers!

PR #256, opened 2021-12-09
SOMD analysis without subsampling

annamherz:
Concerning a transformation in SOMD (ejm31 to ejm42 tyk2 - https://github.com/michellab/BioSimSpace/pull/244#issuecomment-985444208 ) the analysis resulted in a large error because subsampling resulted in less than 50 samples for the bound leg.
When this transformation was rerun, the free energy analysis subsampling still resulted in less than 50 samples for the bound leg and even more vastly wrong results. With the subsampling turned off then for that leg, the results are much better (the experimental is -0.24 kcal/mol)

  | result (kcal/mol) | error
-- | -- | --
subsampling on and resulting in less than 50 samples | -14.7274 | 1.2025
subsampling off if less than 50 samples for that leg | -0.201 | 0.0727

So I think this code change may be a way to avoid that automatically without manually needing to check the output files?

### DIFF
--- a/python/BioSimSpace/FreeEnergy/_relative.py
+++ b/python/BioSimSpace/FreeEnergy/_relative.py
@@ -531,6 +531,13 @@ class Relative():
         if proc.returncode != 0:
             raise _AnalysisError("SOMD free-energy analysis failed!")
 
+        #rerun without subsampling if the subsampling has resulted in less than 50 samples
+        with open("%s/mbar.txt" % work_dir) as file:
+            for line in file:
+                if "#WARNING SUBSAMPLING ENERGIES RESULTED IN LESS THAN 50 SAMPLES" in line:
+                    command = "%s mbar -i %s/lambda_*/simfile.dat -o %s/mbar.txt --overlap" % (_analyse_freenrg, work_dir, work_dir)
+                    proc = _subprocess.run(command, shell=True, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
+        
         # Initialise list to hold the data.
         data = []
 
--- a/python/BioSimSpace/FreeEnergy/_relative.py
+++ b/python/BioSimSpace/FreeEnergy/_relative.py
@@ -535,8 +535,12 @@ class Relative():
         with open("%s/mbar.txt" % work_dir) as file:
             for line in file:
                 if "#WARNING SUBSAMPLING ENERGIES RESULTED IN LESS THAN 50 SAMPLES" in line:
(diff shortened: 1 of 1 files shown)


lohedges:
Thanks @annamherz, this is a good suggestion. Just adding a few suggestions.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -531,6 +531,13 @@ def _analyse_somd(work_dir=None):
         if proc.returncode != 0:
             raise _AnalysisError("SOMD free-energy analysis failed!")
 
+        #rerun without subsampling if the subsampling has resulted in less than 50 samples
+        with open("%s/mbar.txt" % work_dir) as file:
+            for line in file:
+                if "#WARNING SUBSAMPLING ENERGIES RESULTED IN LESS THAN 50 SAMPLES" in line:
+                    command = "%s mbar -i %s/lambda_*/simfile.dat -o %s/mbar.txt --overlap" % (_analyse_freenrg, work_dir, work_dir)
+                    proc = _subprocess.run(command, shell=True, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
+
```
lohedges, line 540: I'd duplicate the check of the process return code from above in case anything else happened to go wrong re-running the command without subsampling. At present, it would just proceed directly to the analysis, perhaps using the original output file if the second command failed. Also put a `break` statement after the return code check so that you skip looping over the rest of the file.

I'm not sure if it's 100% necessasry, but it might also be worth informing the user that their sampling is limited when this fallback is triggered?

lohedges: Thanks, @annamherz. This looks good.

lohedges (approved):

python/BioSimSpace/FreeEnergy/_relative.py
annamherz: Okay, thank you! I can't really think of a good place to include it other than as a print statement, as the function otherwise only returns the data and overlap for that one leg?

kexul:
Hi @annamherz , @lohedges , is it always recommended to turn off subsampling when less than 50 samples were used? In my experiment, only the TI result is affected by this parameter, the mbar result stays unchanged to some extent.  
 
[out.zip](https://github.com/michellab/BioSimSpace/files/7690578/out.zip)

lohedges:
Thanks, I'll take a look. I'm not sure what the recommended best practice is. I imagine the change is quite dependent on the nature of the perturbation and the stability of the sampling, e.g. if fluctuations are small then each sub-sample will be representative.

PR #262, opened 2021-12-31
typos fix

kexul:
Recently I'm trying to add some functionalities to BioSimSpace, Visual Studio Code complains about the syntax errors quite frequently when navigating the project.  So I just fix them all.


lohedges:
Thanks for these. There are a few things that have actually been incorrectly flagged, but I'll fix them after the merge.

PR #273, opened 2022-01-26
Python improvements and enhancements for the 2022.1.0 release

lohedges:
This pull request adds various Python tweaks and performance enhancements in preparation for the 2022.1.0 release. We have now switched to use of `isinstance` whenever possible and no longer run subprocess with `shell=True` (making using of `shlex.split` to split the existing command-line strings).

Many bottleneck parts of the existing code, e.g. molecular manipulation / combination, hydrogen mass repartitioning, water topology substitution, etc., have now been fully implemented in the Sire C++ layer and wrapped to Python. There is now support for native Python pickling of wrapped Sire objects making the existing `BioSimpace.Stream` sub-package redundant.

Continuous integration has been moved over to GitHub actions. We now perform a conda-forge compliant conda build, uploading the resulting package to our channel on the Anaconda cloud.


lohedges:
Fantastic, the BioSimSpace builds are failing since it is pulling in a different version of the Sire dependencies, i.e. boost, so fails. I really don't understand why this happens, since the version _should_ be pinned automatically during the conda-build of Sire. Looks like I'm going to have to re-work the CI for Sire, explicitly pinning all of the dependency versions once again (what we did before). This is what you are always told not to do when creating a conda recipe!

These are the appropriate sections from one of the Linux conda recipes for Sire, i.e. what conda-build creates:

```
host:
libgcc_mutex 0.1 conda_forge
...
    - boost 1.77.0 py38h2b96118_0
    - boost-cpp 1.77.0 h359cf19_1
...

run:
...
    - boost
...
```

As you can see, boost is unpinned as a run dependency, hence the issue. From my understanding, the version used in the host section should be passed through to the run requirement. This is what has happened for other dependencies that we've built against, e.g. qt, tbb, etc.

lohedges:
I'm guessing that conda-forge have their own pinning magic that works on top of conda-build somehow, so we'll need to do something similar in our own build. I'll just pin _everything_ for now. This is a pain though, since it makes dependency resolution increasingly slow as time goes on and means that you can't use more recent versions of BioSimSpace dependencies that share dependencies, even when they are build against what would be a compatible versions.

lohedges:
Looks like [rdkit](https://github.com/rdkit/conda-rdkit/issues/71) have the exact same problem. I'll look to see if they have a solution suggestion. God I hate conda.

lohedges:
Think I might be able put the following in a `conda_build_config.yaml` in the recipe directory:

```
boost: 1.77
pin_run_as_build:
  boost:
    max_pin: x.x
```

lohedges:
Looks like we'll need to do the same thing for OpenMM too, since libOpenMM is unpinned. I'll test locally and see what I get in the `meta.yaml` output.

chryswoods:
That is so frustrating! I feel your pain

lohedges:
That seems to have fixed things locally, but I'll need to do the same for our own `libcpuid` package too. (The `run_exports` doesn't work for some reason.)

This is the resulting `meta.yaml` for the Sire build:

```
run:
    - __glibc >=2.17,<3.0.a0
    - boost >=1.77.0,<1.78.0a0
    - gsl >=2.7,<2.8.0a0
    - libcblas >=3.9.0,<4.0a0
    - libcpuid
    - libgcc-ng >=11.2.0
    - libnetcdf >=4.8.1,<4.8.2.0a0
    - libstdcxx-ng >=11.2.0
    - openmm >=7.7.0,<7.8.0a0
    - pymbar
    - pyqt
    - python >=3.7,<3.8.0a0
    - python_abi 3.7.* *_cp37m
    - qt >=5.12.9,<5.13.0a0
    - tbb >=2021.5.0
```

Note that `boost` and `openmm` are now pinned to the correct versions. I don't know why this isn't done automatically, or isn't set as a `run_export` by the `boost` and `openmm` recipes.

lohedges:
The annoying thing is that we'll need to manually update the dependencies in the `conda_build_config.yaml` as they change, although perhaps I could automate how this is generated.

lohedges:
Okay, I'm currently rebuilding Sire with the pinning in place, then can retest the BIoSImSpace build against it when it's done.

lohedges:
Okay, the CI is now working. I'll merge over to devel, update the tags, then fix any issue with the documentation CI step.

Woohoo!

PR #292, opened 2022-04-05
Addition of AMBER-specific MCS pruning

msuruzhon:
Hello,

This PR adds two new functions: `_prune_perturbed_constraints()`, which removes all hydrogens that have been perturbed to non-hydrogens from the MCS, and `_prune_crossing_constraints()` which makes sure there are no constrained bonds between dummy atoms and non-dummy atoms and prunes the MCS to achieve that. These two pruning procedures have been added to `matchAtoms()` as an option, and their default values are based on the MD engine, which is another settable parameter in `matchAtoms()`, with the default being `"SOMD"` (thereby preserving the current default behaviour). These prunings have been validated on a variety of systems and result in (and are required for) consistently numerically stable simulations on CUDA at 4 fs with HMR.

We might need to consider promoting this `engine` parameter to the `merge()` function as well. I haven't done that yet but I will do it if you think this makes sense. @lohedges @annamherz @jmichel80 

Many thanks.


lohedges:
Hi @msuruzhon,

Thanks for this, it sounds really helpful. I'll be able to comment on specifics of the code later in the week, but just a few general thoughts/comments for now.

You mention that these modifications are AMBER specific. In addition, they are HMR and PMEMD-CUDA specific too (I guess you wouldn't even try such a simulation on CPU anyway). Are there cases in which they may be applicable to SOMD / GROMACS, or are they related to the specific way in which AMBER implements the alchemical transformation?

In general, we have tried to take the approach that the set-up stages are simulation engine agnostic, with all specialisations of this nature done at the process creation stage, i.e. when an appropriate engine has been selected, hardware detected / chosen, etc. This helps with keeping things interoperable and allows the user to break their workflow into separate nodes, e.g. for setup, running, and analysis, each of which would work interoperably. By pinning something, e.g. the MCS, to an engine early on, you then constrain the engine throughout.

With the above in mind, I'd personally prefer if `matchAtoms` always returns a generic MCS mapping, which could then be specialised for engine/simulation constraints later on. (Perhaps warning the user if when changes have been made.) This might be a bit more tricky, since we'd need to modify the merged topology directly, but could be simplified by storing the MCS and arguments to the `merge` function as attributes of the perturbable molecule, then re-running `merge` with your pruned MCS within the setup stage of `BioSImSpace.Process.Amber`. (Your functions would still live within `BioSImSpace.Align._align.py`, they would just be called elsewhere.)

If the pruning is only needed when HMR is being used, then we could also flag that HMR has been applied as a system attribute so that we know when to apply it.

> We might need to consider promoting this engine parameter to the merge() function as well.

As mentioned above, I'd rather the implementation was moved elsewhere, but to comment on this... We've taken the approach that calling `merge` with no-mapping will generate one for you by calling `matchAtoms` with the default options. If the user wants to specialise things, then they should break things into multiple stages, passing the specific arguments that they require, i.e. first calling `matchAtoms`, then passing the resulting mapping to `merge`.

Cheers.

lohedges:
For example, SOMD does some things to specialise the MCS in the `to_pert_file` function that is called within `BioSImSpace.Process.Somd._setup()`. This modifies certain terms to improve simulation stability.

lohedges:
Thanks for the clarification, that's very helpful.

I'm not too adverse to some redundancy in the name of interoperability, e.g. repeated pruning. In practice these things should be fast in comparison to the length of a simulation so aren't likely to be a bottleneck.

That said, it seems like people are moving towards a model of using BioSimSpace where they want to select the engine up front, rather than have it determined automatically. It is quite hard to cleanly support both of these approaches.

Perhaps we need to have a think about when and where hardware is detected, i.e. should this be done immediately on module load and be fixed, or always set as an input to a node, with auto used as a fallback? This would simplify some of the issues you are describing since the workflow would always be compatible with the selected engine. This gets tricky when working across multiple nodes, though, or re-using existing output in another workflow. This is what we always had in mind, i.e. you could create a set of reusable and interoperable workflow components that could be plugged together in different ways.

I do have plans for a sub-package to allow users to pre-configure the hardware and software setup, along with cluster specifications. This would make things much easier too.

Perhaps we could have the `engine` option to `matchAtoms` for power users, but still have some code within the process objects to still detect cases where the mapping wouldn't work. This could get complicated though.

Thanks for the info on HMR too. This functionality is quite new and I agree that it is important to make its use consistent with the engine's requirement. At present it is implemented using standalone methods on a system object, so you need to know what you are doing, i.e. what engine you will use. As you say, it would perhaps be better to add HMR via a protocol option, or apply it whenever the time step is above a certain size. As you've mentioned, this would lead to redundancy though. (Again, it should be quick compared to the simulation.)

Cheers.

bcossins:
Could keep things engine agnostic but alchemical topology specific, ie.
Single topology (GROMACS and SOMD), dual topology (ABFE), hybrid topology (AMBER RBFE), chimera topology (GROMACS)
?

lohedges:
Something like that might work. I think the issue with doing that in the current setup is that the topology specific mapping is just a dictionary, so you'd also need to pass through what topology the mapping corresponds to so that an appropriate engine is selected. (I guess it could be stored as an attribute of the merged molecule if it was also passed to `merge`.)

It's often difficult to work out where certain things should be done and you can quite easily run into chicken-and-egg situations, i.e. which thing needs to know what first, etc.

I think it'll take a bit of thought to work out the best way of doing things. @msuruzhon  and @annamherz 's input is invaluable, since they are at the front line of actually having to use this.

lohedges:
Yes, I agree with this approach. I think one of the difficulties with BioSimSpace is that we always wanted a simple and robust API that we could support for the long term, e.g. having long lived nodes that wouldn't break. We never had a pre-alpha, alpha, ... approach where we had a volatile API where we could iron out wrinkles. This has made it difficult to rejig things as problems have arisen while keeping the API consistent. It's the challenge of having interoperability a precursor, rather than the end goal, i.e. in your suggestion we could support multiple (somewhat) independent engine appropriate FEP implementations, then wrap these (as far as possible) behind a consistent interoperable interface once they are robust and working. (Rather than having the interface first, then trying to fit the different engine specific implementations into it. Which is particularly difficult when you are unfamiliar with them at the outset.)

I think it would be good to have a discussion about the broader development approach at some point. It would be great to try to keep our lofty interoperable ideals, but perhaps we need to be more flexible at the development stage to allow developers and users to flesh out ideas, etc.

Cheers.

lohedges:
@annamherz: I'm happy for this to be merged across to your feature branch. At this stage, it's important that you can play around with / test the changes. We can think about aligning things with the existing interface in `devel` when we get to that point.

I'm not sure if you have permissions to merge. If not, let me know and I'll do it for you.

Cheers.

annamherz:
Okay, I'll go ahead and merge for now!

jmichel80:
Congrats for this milestone !

msuruzhon:
Thanks for the feedback @lohedges , ideally I would also like to keep the MCS part engine-agnostic but looking at how every engine has its own very specific way of defining the topology/soft-core potential, I don't think it is possible to do that without sacrificing computational efficiency during the simulation stage - a mapping that is good for GROMACS is not necessarily good for AMBER in the general case. So it would be good to have reasonable defaults while at the same time allowing for flexibility. That being said, I am not averse to having a separate function that does the pruning so one could then chain `matchAtoms()`, `prune()` and `merge()` in a functional way.

Do you have any rough estimate of when this is going to be merged into `devel`? I am conscious that we again run the risk of the branch diverging and as of now this makes using the latest pushes to `devel` challenging.

Cheers.

lohedges:
Hi @msuruzhon.

> Do you have any rough estimate of when this is going to be merged into `devel`? I

I don't have a timeline, I'm afraid. It's challenging for me to look at something this significant while not being full time on the project. As mentioned in the initial merge request, I'd like to incorporate the Config generator in a more consistent way, i.e. it's own sub-package, and we also need to work out how to implement the new FreeEnergy protocols that you've added, since they wouldn't work with all engines, e.g. SOMD.

I have been working on porting across your Config work in my own local branch, and should be able to merge this in at some point. I think it might be easier if I do things in stages, since I like  to think about a _single_ feature in isolation and work that in, rather than everything at once.

@annamherz Has merged the updates from `devel` into your original branch, but I am aware of some inconsistencies from the merge that would also need to be addressed. (The things I spotted were fairly minor, but didn't go through things with a fine-toothed comb.) As long as you both keep regularly pulling across updates from `devel` then things shouldn't diverge. This will also make it easier for me to merge things in bit by bit.

>  I don't think it is possible to do that without sacrificing computational efficiency during the simulation stage - a mapping that is good for GROMACS is not necessarily good for AMBER in the general case. So it would be good to have reasonable defaults while at the same time allowing for flexibility. That being said, I am not averse to having a separate function that does the pruning so one could then chain matchAtoms(), prune() and merge() in a functional way.

Is it always the case that the mapping for AMBER is simply a pruned version of the general mapping, or in some cases would an _entirely_ different mapping be preferable? If it's the former, I still think we could just move the pruning into the engine specific part, i.e. the process. setup As mentioned, the way the mapping is actually applied is slightly different for SOMD / GROMACS too, which is done when the perturbation file is generated, or the merged topology is written to file. (In this case nothing is pruned, just terms modified, e.g. for those involving dummies, to improve stability.) If we really need a different mapping at the `matchAtoms` stage, then I don't see how we can do things without making the FEP setup stage engine specific. If this is the case, then I think it would be necessary to rethink the whole mapping --> merge --> FEP simulation setup approach in BioSImSpace, or think harder about when and where we want to implement engine specific functionality.

Cheers.

lohedges:
> However, looking at how the default mapping is the "true" maximum common substructure, I guess we should be always able to get away with pruning because there is no way to make the MCS bigger?

Yes, that's what I was thinking, i.e. `matchAtoms` returns the _true_ MCS, then the process specific setup will tweak or prune this as necessary for the engine's specific requirements. (Obviously it might make sense to return a different mapping for a different type of FEP simulation, though.)

I was only wondering if there are other subtleties of the AMBER-specific mapping, rather than it just being a pruning of the general MCS.

Cheers.

PR #295, opened 2022-04-25
Use pexpect to run trajconv

kexul:
Dear BioSimSpace developers: 
     Recently I found that BioSimSpace always failed to getSystem() for my system, here are the code and [input files](https://github.com/michellab/BioSimSpace/files/8552976/temp.zip): 

```python 
def run(sys, protocol, workdir):
    process = BSS.Process.Gromacs(sys, protocol, ignore_warnings=True, work_dir=workdir)

    process.addArgs({'-ntmpi':1, '-ntomp':'8', '-gpu_id':0})
    # Start the process in the background.
    process.start()

    # Wait for the process to finish.
    process.wait()

    # Get the minimised molecular system.
    minimised = process.getSystem()
    return minimised

inp = 'protein.inpcrd'
top = 'protein.prmtop'

mol = BSS.IO.readMolecules([inp, top])
sys = BSS.Solvent.tip3p(mol, [8*BSS.Units.Length.nanometer]*3, is_neutral=True, work_dir='aa')

mini = BSS.Protocol.Minimisation(steps=10000)
minied = run(sys, mini, 'bb')

nvt_short = BSS.Protocol.Equilibration(runtime=50*BSS.Units.Time.picosecond, temperature_start=0*BSS.Units.Temperature.kelvin,
                                        temperature_end=300*BSS.Units.Temperature.kelvin,
                                        )
nvtd = run(minied, nvt_short, 'cc')

```

Output: 
```
Failed to extract trajectory frame with trjconv. Try running 'getSystem' again.
```

After some debugging, I found it might be failing here, BioSimSpace tries to read the `frame.gro` while the frame extract process is ongoing: 
https://github.com/michellab/BioSimSpace/blob/92bc3af971905309d6f09b249a1972309a6f0874/python/BioSimSpace/Process/_gromacs.py#L2478-L2479

This commit use pexpect to run trajconv, expect 'Gromacs reminds you' to make sure trajconv is finished.

### DIFF
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -38,6 +38,7 @@ import shlex as _shlex
 import subprocess as _subprocess
 import timeit as _timeit
 import warnings as _warnings
+import pexpect
 
 from Sire import Base as _SireBase
 from Sire import IO as _SireIO
@@ -2460,19 +2461,12 @@ class Gromacs(_process.Process):
                 command = "%s trjconv -f %s -s %s -dump %f -pbc mol -o frame.gro" \
                     % (self._exe, self._traj_file, self._tpr_file, time.picoseconds().value())
 
-                # Run the command as a pipeline.
-                proc_echo = _subprocess.Popen(["echo", "0"], shell=False, stdout=_subprocess.PIPE)
-                proc = _subprocess.Popen(_shlex.split(command), shell=False,
-                    stdin=proc_echo.stdout, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
-                proc_echo.stdout.close()
-
-                # For some reason, this doesn't always work the first time it's run
-                # as a subprocess, so try again if frame.gro isn't found.
-                if not _os.path.isfile("frame.gro"):
-                    proc_echo = _subprocess.Popen(["echo", "0"], shell=False, stdout=_subprocess.PIPE)
-                    proc = _subprocess.Popen(_shlex.split(command), shell=False,
-                        stdin=proc_echo.stdout, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
-                    proc_echo.stdout.close()
+                proc = pexpect.spawn(command, encoding='utf-8')
+                proc.expect('Select a group:')
+                proc.sendline('0')
+                proc.sendline()
+                proc.expect('GROMACS reminds you: ')
+                proc.close()
 
                 # Read the frame file.
                 new_system = _IO.readMolecules(["frame.gro", self._top_file],


lohedges:
Thanks for this, it explains why I was intermittently getting no output from `getSystem`, while the exact same commands worked when run directly on the command-line. I added a workaround to re-run if the original command fails, but this seems more robust. 

I'll check in more detail when I'm back later in the week and will let you know if I have any questions.

(For reference, this has only been an issue since switching the way the I used `subprocess`, by no longer supporting `shell=True`.)

Cheers.

kexul:
> I'll check in more detail when I'm back later in the week and will let you know if I have any questions.

Thanks 🤗 

 I just tested `shell=True`, it does not work for my system either.

lohedges:
It won't with the new approach that involves splitting it into two processes with a pipe. Previously it was just a single command run with `shell=True`, which should wait for the process to exit before returning. (Something strange could still be going on, though.)

kexul:
Hi @lohedges , I just tested the following code: 
```python 
command = "echo 0 | %s trjconv -f %s -s %s -dump %f -pbc mol -o frame.gro" \
                    % (self._exe, self._traj_file, self._tpr_file, time.picoseconds().magnitude())
proc = _subprocess.Popen(command, shell=True,
                     stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
```

The same error occurred: 
```
UserWarning: Failed to extract trajectory frame with trjconv. Try running 'getSystem' again.
  _warnings.warn("Failed to extract trajectory frame with trjconv. "
```

lohedges:
Hi again,

Looking back at my original code it looks like I've forgotten to wait for the process to finish before continuing. Could you try with the following and let me know if it works for you?

```python
                # Use trjconv to get the frame closest to the current simulation time.
                command = "%s trjconv -f %s -s %s -dump %f -pbc mol -o frame.gro" \
                    % (self._exe, self._traj_file, self._tpr_file, time.picoseconds().value())

                # Run the command as a pipeline.
                proc_echo = _subprocess.Popen(["echo", "0"], shell=False, stdout=_subprocess.PIPE)
                proc = _subprocess.Popen(_shlex.split(command), shell=False,
                    stdin=proc_echo.stdout, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
                proc.wait() # Wait for process to finish!
                proc_echo.stdout.close()

                # Comment out repeat attempt.
                #if not _os.path.isfile("frame.gro"):
                    #proc_echo = _subprocess.Popen(["echo", "0"], shell=False, stdout=_subprocess.PIPE)
                    #proc = _subprocess.Popen(_shlex.split(command), shell=False,
                    #    stdin=proc_echo.stdout, stdout=_subprocess.PIPE, stderr=_subprocess.PIPE)
                    #proc_echo.stdout.close()
```

(Previously I was using a single subprocess command that would automatically wait. Here I've split it into two `Popen` commands to allow a pipe, but forgot that I now need to wait for the latter to finish.)

This works for me using the example input that your provided. This solution would avoid the need for `pexpect` (a third-party package). I would also need to update similar code in the `Solvate` sub-package, although that hasn't caused issues as of yet, presumably since the command finishes quickly.

Cheers.

kexul:
Thanks, this works for me too! 🥂

lohedges:
Brilliant. I'll push my updates now.

PR #301, opened 2022-05-09
Feature sandpit: Support for experimental features alongside devel

lohedges:
This pull requests implements a _sandpit_ within BioSImSpace to allow support for experimental features along side the development build of BioSimSpace. To import a specific sandpit one can do, e.g.:

```python
```

The sandpit requires use of _relative_ import statements to ensure that packages/modules are imported from the correct sandpit, rather than the base development package. For example, use:

```python
from . import Protocol as _Protocol
```

rather than:

```python
import BioSimSpace.Protocol as _Protocol
```

(Obviously the number of `.` characters needed will depend on which directory level you are in, i.e. you'd need to do `from .. import Protocol as _Protocol` if you were one level down.)

For users of a particular sandpit, please fork this repository and work within your own sandpit space. Once you are happy with edits, raise a pull request against our development branch to merge in updates to your sandpit. We will work to port updates from devel into your sandpit. This will take the form of a named feature branch, e.g. `feature_sandpit_exscientia`, from which we will create a pull request against devel. We will assign an appropriate reviewer for the pull request to ensure that you get the chance to review and test any changes before merging.

Cheers.


lohedges:
Looks like I need to add a separate dunder init to the sandpit directory, since nothing is imported when using:

```python
```

i.e. `BSS` is empty.

I'll try to fix this later.

lohedges:
Okay, it was working, I was just trying on `devel`, where the dunder init isn't present. I've now added a page to the website to explain the sandpit and how it should be used.

msuruzhon:
Hi @lohedges, what would be the best way to change all imports to the ones in the `Sandpit`? Will I have to replace all `BioSimSpace.XXXXX` imports with `BioSimSpace.Sandpit.Exscientia.XXXXX` or is there a more straightforward way to do that?

lohedges:
I've already done this for you.

msuruzhon:
Thanks @lohedges, I meant all imports from outside of `BioSimSpace`.

lohedges:
Ah, I see, sorry about that. You mean that you are importing specific a sub-set of BioSimSpace functionality in external scripts?, I think the easiest approach would just be to do what's suggested at the start, i.e. start any script with something like:

```python
```

If you just want to use specific sub-packages, then you would need to do something like you suggest, e.g:

```python
import BioSimpace.Sandpit.Exscientia.FreeEnergy as FreeEnergy
```

msuruzhon (approved):
Thanks @lohedges I tested this new branch and all of my tests pass so there don't seem to be any issues with the paths.

lohedges:
Great, thanks for confirming all is okay. I'll merge in tomorrow.

I guess you'll also need to update paths in your wrappers to patch certain files, i.e. patching things within the sandpit.

lohedges:
Well, remarkably _all_ builds failed following the merge, despite _all_ checks passing in the PR. This is with the same test failure, which doesn't fail locally. It _does_ intermittently fail in the CI, but I've _never_ seen it fail for _all_ builds at the same time. Baffling!

msuruzhon:
Yes, it is very strange why it builds fine for `python=3.7` and `python=3.9` but complains about missing files in `python=3.8`...

lohedges:
Just re-run and it now passes on all but two builds. I've no idea why this is so flaky. I think I might just delete the test for now, or make it something that is only run locally.

lohedges:
Third time lucky :man_shrugging:

PR #308, opened 2022-05-13
Allows the tests available in test to be in test/Sandpit/Exscientia

xiki-tempula:
Copy the tests from the `test` to `test/Sandpit/Exscientia`

- Change the import in `test/Sandpit/Exscientia`
- Fix the test to ensure that it runs
- Change the default pytest import to `--import-mode=importlib`


lohedges:
Thanks for this. I'll just check with @chryswoods as to whether we want a complete clone of the tests within `test/Sandpit/Exscientia`. If so, then you'd need to update _all_ imports within the test scripts to use `BioSimSpace.Sandpit.Exscientia`, rather than the main development code.

I was originally thinking that the directory would have the same directory structure as the default tests, but only include files related to additions that you have made. However, it would probably make sense to duplicate to ensure that additions are consistent with devel as you are making them.

Cheers.

xiki-tempula:
Thanks for the comments/

> If so, then you'd need to update all imports within the test scripts to use BioSimSpace.Sandpit.Exscientia, rather than the main development code.

I think this what I'm doing here where I change all the tests in test/Sandpit/Exscientia to import from `BioSimSpace.Sandpit.Exscientia`

lohedges:
Yes, but I don't import changes in all files. For example, in `test/Sandpit/Exscientia/Align/test_align.py` you have updated to:

```python
```

but in other files, such as `test/Sandpit/Exscientia/Gateway/test_file.py` you still have

```python
from BioSimSpace.Gateway import File
```

This would need to be:

```python
from BioSimSpace.Sandpit.Exscientia.Gateway import File
```

While the two files might be identical for now, this would ensure that the test still tracked your sandpit if changes were made at a later stage.

Cheers.

xiki-tempula:
@lohedges Oh, thanks. I have only replaced the `import BioSimSpace as BSS`

xiki-tempula:
@lohedges I have changed the relevant files now.

lohedges:
Hmm, I still see quite a few files that haven't been updated, such as this one: https://github.com/Exscientia/BioSimSpace/blob/feature_test/test/Types/test_types.py

I'd open _all_ python files within `test/Sandpit/Exscientia` and update them one-by-one.

lohedges:
Just a note that it's useful to have informative commit messages. It makes it much more useful to work out what was going on when going back through the commit history, e.g. when trying to track down a bug. Something like "Update import paths to use Exscientia sandpit" would be fine.

Cheers.

xiki-tempula:
@lohedges Sorry, I usually change the commit message when squash and merge. I have checked all of the test files and can verify that they have all been changed.

lohedges:
Great, thanks for that. I've just checked-out your branch locally and am just checking that the tests pass for me.

xiki-tempula:
@lohedges Sorry if this has already been discussed. I wonder why don't we run test as part of the CI?

lohedges:
Ah yes, good point. I can always squash when merging. I generally don't do this unless the commits are very small since it makes finding bugs trickier. (In this case it wouldn't be a problem, though.)

> @lohedges Sorry if this has already been discussed. I wonder why don't we run test as part of the CI?

Tests are run as part of the CI. I just always like to check locally before pushing since it's a waste of cloud resources if there's already a known failure.

Having run the tests locally I see three failures.

```
test/Sandpit/Exscientia/IO/test_openff_gromacs.py::test_molecule_combine FAILED
test/Sandpit/Exscientia/Process/test_single_point_energy.py::test_amber_gromacs FAILED
test/Sandpit/Exscientia/Process/test_single_point_energy.py::test_amber_gromacs_triclinic FAILED
```
The error is:

```
E               RuntimeError: Unable to generate GROMACS binary run input file.
E
E               'gmx grompp' reported the following errors:
E               ERROR 1 [file /tmp/tmp7n3ze051/gromacs.mdp, line 3]:
E                 Parameter "nsteps" doubly defined
```

These fail because the `setConfig` method on the process is inconsistent with the `ConfigFactory` introduced by @msuruzhon, i.e. the old config isn't being overwritten and keys with duplicate values are somehow being included twice.

xiki-tempula:
@lohedges I have had a check and I don't have amber installed so all these tests were skipped.

lohedges:
No problem. Yes, some of the tests require external packages so they are skipped if not installed. We don't install AmberTools or GROMACS as dependencies via conda as the user will likely have custom installations optimised for their HPC resources. The conda packages also used to to be incompatible with the BioSimSpace environment too. Going forward, it might be worth me checking whether they can be installed during the CI for testing purposes. (So these tests can be run remotely too.)

lohedges:
The issue isn't the `ConfigFactory`, rather a the `nstlog` option has also been added to the `Minimisation` protocol, so `nsteps` is no longer at entry 1 in the config. The tests pass with the following diff:

```
diff --git a/test/Sandpit/Exscientia/IO/test_openff_gromacs.py b/test/Sandpit/Exscientia/IO/test_openff_gromacs.py
index 77c5efc5..3f63f7a5 100644
--- a/test/Sandpit/Exscientia/IO/test_openff_gromacs.py
+++ b/test/Sandpit/Exscientia/IO/test_openff_gromacs.py
@@ -47,7 +47,7 @@ def test_molecule_combine():

     # Modify the GROMACS configuration to run zero steps.
     config = p0.getConfig()
-    config[1] = "nsteps = 0"
+    config[2] = "nsteps = 0"
     p0.setConfig(config)
     p1.setConfig(config)
     p01.setConfig(config)
diff --git a/test/Sandpit/Exscientia/Process/test_single_point_energy.py b/test/Sandpit/Exscientia/Process/test_single_point_energy.py
index 7402af33..b210f13d 100644
--- a/test/Sandpit/Exscientia/Process/test_single_point_energy.py
+++ b/test/Sandpit/Exscientia/Process/test_single_point_energy.py
@@ -37,7 +37,7 @@ def test_amber_gromacs(system):

     # Modify the GROMACS configuration to run zero steps.
     config = process_gmx.getConfig()
-    config[1] = "nsteps = 0"
+    config[2] = "nsteps = 0"
     process_gmx.setConfig(config)

     # Run the AMBER process and wait for it to finish.
@@ -84,7 +84,7 @@ def test_amber_gromacs_triclinic(system):

     # Modify the GROMACS configuration to run zero steps.
     config = process_gmx.getConfig()
-    config[1] = "nsteps = 0"
+    config[2] = "nsteps = 0"
     process_gmx.setConfig(config)

     # Run the AMBER process and wait for it to finish
```
Could you patch the files and update?

PR #310, opened 2022-05-13
Add method to get index of lambda value in lambda array. [closes #309]

### LINKED ISSUE #309: Allow the BSS.Protocol.FreeEnergy to have the method getLambdaIndex
**Is your feature request related to a problem? Please describe.**
So usually, one need Lambda, LambdaValues and LambdaIndex to run free energy calculations.
Currently there isn't a getLambdaIndex method and this is done via LambdaValues.index(Lambda) https://github.com/michellab/BioSimSpace/blob/devel/python/BioSimSpace/Process/_gromacs.py#L565

Supporting this method would mean that we could change something within the BSS.Protocol.FreeEnergy (like use pandas.DataFrame for the exs stuff) without changing the public API.

**Describe the solution you'd like**
Allow BSS.Protocol.FreeEnergy to have the method getLambdaIndex().

lohedges:
This PR addresses the feature request in #309.

### DIFF
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -557,12 +557,11 @@ class Gromacs(_process.Process):
                     % self._protocol.getPressure().bar().value())
                 config.append("compressibility = 4.5e-5")   # Compressibility of water.
 
-            # Extract the lambda value and array.
-            lam = self._protocol.getLambda()
+            # Extract the lambda array.
             lam_vals = self._protocol.getLambdaValues()
 
             # Determine the index of the lambda value.
-            idx = lam_vals.index(lam)
+            idx = self._protocol.getLambdaIndex()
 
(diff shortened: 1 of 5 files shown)


xiki-tempula:
@lohedges 

> This should have been merged into our devel, then your devel merged with ours, then your feature branch merged with your devel, then your feature branch PR'ed back into your devel.

I think I will do this when we have a complete ABFE protocol as we have underwent several versions of Protocol when I noticed that other classes might need Protocol in a way that I hadn't thought about.

msuruzhon:
Sorry @lohedges I didn't realise I was tagged as a reviewer on this PR. Let's merge it to your `devel` sandpit so we could sync ours with yours. If there are any conflicts in this instance we will handle them.

Once we have something working then we will submit a PR to your `devel` sandpit from ours. Is this okay?

Many thanks.

Issue #326, opened 2022-05-24
Support pow operator for all unit-based types


xiki-tempula:
So I would like to represent the force constant for a harmonic angle restraint, which would be `kcal_per_mol / radian ** 2`

I tried to create the force constant as
```
from BioSimSpace.Units.Angle import radian
from BioSimSpace.Units.Energy import kcal_per_mol
fc = 1 * kcal_per_mol / radian ** 2
```
However, I would get
```
  File "<ipython-input-4-b78feb2eada7>", line 1, in <module>
    radian**2
TypeError: unsupported operand type(s) for ** or pow(): 'Angle' and 'int'
```

`radian * radian` works though.

lohedges:
Yes, if doing `radian * radian` it will create a `GeneralUnit`. I just need to do the same thing for the `pow` operator. (The `__mul__` and `__rmul__` operators already do this.)

xiki-tempula:
@lohedges When you are going to do it. 
Do you mind making it such that
`isinstance(obj, BSS.Types.Angle ** 2)` will work as well? 
I know that one could check via dimensions() but it would be better if this is possible.
Or if this is not possible something like `obj.isType(BSS.Types.Angle ** 2)` would be good as well.

PR #327, opened 2022-05-24
Support pow operator for all unit-based types. [ref #326]

lohedges:
This PR addresses the feature request in https://github.com/michellab/BioSimSpace/issues/326, i.e. copying the updates from devel into the Exscientia sandpit. There were no conflicts to resolve and all unit tests pass.


msuruzhon (changes requested):
Thanks @lohedges, looks good to me but it would be nice to add a simple unit test which verifies this functionality.

Cheers.

PR #328, opened 2022-05-25
Allow BSS to do ligand in water decoupling calculations

xiki-tempula:
Description
This PR adds allows one to decouple a molecule in solvent to compute the desolvation free energy.

Type of change
Bug fix (non-breaking change which fixes an issue)

- Fix the unit test to have them running fine on GMX 2022.1
- Change the maxwarn to use the Gromacs unlimited option

New feature (non-breaking change which adds functionality)

- Add the Github actions such that the CI will run all the unit tests
- Use pd.DataFrame and pd.Series internally in the BSS.Protocol._FreeEnergyMixin such that the more than one lambda state is supported.
- Add setLambda method to BSS.protocol.free_energy_mixin
- When supplying a Protocol with multiple lambda state, the BSS.Protocol._config could write multiple stage lambda into Gromacs mdp files 
- Change the default mdp options to better suit the gmx-2022
- Allow BSS.FreeEnergy.relative to use the new lambda config 
- Write test to ensure that the mdp file has been modified
- Allow the BSS.FreeEnergy.Relative to handle decoupled molecule
- Allow the BSS.Protocol.Config to edit the mdp file such that decoupled molecule could be written
- Add new method BSS.Align.Decouple to make molecule as being decoupled
- Allow BSS.mol and BSS.system to handle decoupled state properly

Changes (breaking change which adds functionality)

- Change the folder being created to use the index of the lambda (e.g. lambda_0, lambda_1 ... lambda_32)

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/Align/_decouple.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/Align/_decouple.py
@@ -1,0 +1,95 @@
+import warnings
+
+from Sire import Base as _SireBase
+
+from .._SireWrappers import Molecule as _Molecule
+from .._Exceptions import IncompatibleError as _IncompatibleError
+
+
+def decouple(molecule, property_map0=None, property_map1=None, intramol=True):
+    """Make the molecule as being decoupled, where the interactions with the
+    rest of the environment are removed, or annihilate, where the interactions
+    within the molecule are removed as well (choose this mode with
+    intramol=False).
+
+        Parameters
+        ----------
+
+        molecule : BioSimSpace._SireWrappers.Molecule
+            The molecule to be decoupled or annihilated.
+
+        property_map0 : dict
+            A dictionary that maps "properties" in this molecule to their
+            user defined values at the start of the transformation. This allows
+            user to selectively turn on or off the charge or the vdw
+            interactions e.g. { "charge" : True, "LJ" : True}
+
+        property_map1 : dict
+            A dictionary that maps "properties" in this molecule to their
+            user defined values at the end of the transformation. This allows
+            user to selectively turn on or off the charge or the vdw
+            interactions e.g. { "charge" : False, "LJ" : False}
+
+        intramol : bool
+            Whether to remove the intra-molecule forces, thereby annihilate the
+            molecule instead of decouple it.
+
+        Returns
+        -------
+
+        decoupled : Sire.Mol.Molecule
+            The molecule marked as being decoupled.
+    """
+    # Validate input.
+
+    if not isinstance(molecule, _Molecule):
+        raise TypeError("'molecule' must be of type 'BioSimSpace._SireWrappers.Molecule'")
+
+    # Cannot decouple a perturbable molecule.
+    if molecule.isDecoupled():
+        raise _IncompatibleError("'molecule' has already been decoupled!")
+
+    if property_map0 is None and property_map1 is None:
+        property_map0 = {"charge": True, "LJ": True}
+        property_map1 = {"charge": False, "LJ": False}
+    if property_map0 is None:
+        property_map0 = {}
+    if property_map1 is None:
+        property_map1 = {}
+
+    # Check the keys in the property_map
+    for property_map in (property_map0, property_map1):
+        if not isinstance(property_map, dict):
+            raise TypeError("'property_map' must be of type 'dict'")
+        for key in property_map:
+            if not key in ('charge', 'LJ'):
+                warnings.warn(f'Key {key} not supported for decouple, will be '
+                              f'ignored. The recognised keys are '
+                              f'("charge", "LJ")')
+
+    if not isinstance(intramol, bool):
+        raise TypeError("'intramol' must be of type 'bool'")
+
+    # Create a copy of this molecule.
+    mol = _Molecule(molecule)
+    mol_sire = mol._sire_object
+
+    # Edit the molecule
+    mol_edit = mol_sire.edit()
+
+    # Set the start and the end state
+    mol_edit.setProperty("charge0", _SireBase.wrap(property_map0.get("charge",True)))
+    mol_edit.setProperty("charge1", _SireBase.wrap(property_map1.get("charge", True)))
+    mol_edit.setProperty("LJ0", _SireBase.wrap(property_map0.get("LJ", True)))
+    mol_edit.setProperty("LJ1", _SireBase.wrap(property_map1.get("LJ", True)))
+
+    mol_edit.setProperty("annihilated", _SireBase.wrap(not intramol))
+
+    # Flag that this molecule is decoupled.
+    mol_edit.setProperty("is_decoupled", _SireBase.wrap(True))
+
+    # Update the Sire molecule object of the new molecule.
+    mol._sire_object = mol_edit.commit()
+
+    return mol
+
--- a/test/Sandpit/Exscientia/Align/test_decouple.py
+++ b/test/Sandpit/Exscientia/Align/test_decouple.py
@@ -1,0 +1,65 @@
+import pytest
+
+import BioSimSpace.Sandpit.Exscientia as BSS
+from BioSimSpace.Sandpit.Exscientia.Align._decouple import decouple
+
+
+@pytest.fixture()
+def mol():
+    # Benzene.
+    mol = BSS.Parameters.openff_unconstrained_2_0_0(
+        "c1ccccc1").getMolecule()
+    return mol
+
+def test_sanity(mol):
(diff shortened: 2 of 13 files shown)


lohedges:
Thanks, I'll add some specific comments to the code.

test/Sandpit/Exscientia/Align/test_decouple.py
```diff
@@ -0,0 +1,65 @@
+import pytest
+
+import BioSimSpace.Sandpit.Exscientia as BSS
+from BioSimSpace.Sandpit.Exscientia.Align._decouple import decouple
+
+
+@pytest.fixture()
+def mol():
+    # Benzene.
+    mol = BSS.Parameters.openff_unconstrained_2_0_0(
```
lohedges, line 10: Parameterising with OpenFF requires that AmberTools is installed (it requires `antechamber`). Since AmberTools is optional, please add a guard so that the test can be skipped if it's not installed. See [here](https://github.com/michellab/BioSimSpace/blob/devel/test/Process/test_amber.py) for an example. (Since it only requires `antechamber`, which must be in the `PATH`, you could instead check for that.)

test/Sandpit/Exscientia/Free_Energy/test_relative.py
```diff
@@ -0,0 +1,59 @@
+import pathlib
+
+import pytest
+import pandas as pd
+import numpy as np
+from alchemlyb.parsing.gmx import extract_u_nk
```
lohedges, line 6: `alchemlyb` isn't part of the development BioSimSpace dependencies, so this would break our CI. Please wrap the import in a `try/except` block and skip the tests if it can't be imported.

test/Sandpit/Exscientia/Free_Energy/test_relative.py
```diff
@@ -0,0 +1,59 @@
...
+import BioSimSpace.Sandpit.Exscientia as BSS
+from BioSimSpace.Sandpit.Exscientia.Protocol import FreeEnergyEquilibration
+from BioSimSpace.Sandpit.Exscientia.Align._decouple import decouple
+from BioSimSpace.Sandpit.Exscientia import Types as _Types
+
+class Test_gmx_ABFE():
+    @staticmethod
+    @pytest.fixture(scope='class')
+    def freenrg():
+        m = BSS.Parameters.openff_unconstrained_2_0_0("c1ccccc1").getMolecule()
```
lohedges, line 17: Similar issue with requiring `antechamber` here.

test/Sandpit/Exscientia/Free_Energy/test_relative.py
```diff
@@ -0,0 +1,59 @@
...
+        decouple_m = decouple(m,
+                              property_map0={"charge": True, "LJ": True},
+                              property_map1={"charge": False, "LJ": False},
+                              intramol=False)
+        solvated = BSS.Solvent.tip3p(molecule=decouple_m,
```
lohedges, line 22: `BSS.Sovlent` requires that GROMACS is installed, so the test should also be skipped when it isn't installed. See [here](https://github.com/michellab/BioSimSpace/blob/devel/test/Process/test_gromacs.py) for an example.

test/Sandpit/Exscientia/Protocol/test_config.py
```diff
@@ -0,0 +1,113 @@
...
+            steps=10000, perturbation_type="full")
+        freenrg = BSS.FreeEnergy.Relative(system, protocol, engine='GROMACS', )
+        with open(f"{freenrg._work_dir}/lambda_6/gromacs.mdp", 'r') as f:
+            mdp_text = f.read()
+            assert 'bonded-lambdas = 0.0 0.25 0.5 0.75 1.0 1.0 1.0' in mdp_text
+            assert 'coul-lambdas = 0.0 0.0 0.0 0.5 1.0 1.0 1.0' in mdp_text
+            assert 'vdw-lambdas = 0.0 0.0 0.0 0.0 0.0 0.5 1.0' in mdp_text
+            assert 'init-lambda-state = 6' in mdp_text
+
+class TestGromacsABFE():
```
lohedges, line 71: These tests also need to be skipped if `antechamber` or GROMACS aren't installed.

python/BioSimSpace/Sandpit/Exscientia/Align/_decouple.py
```diff
@@ -0,0 +1,95 @@
...
+    within the molecule are removed as well (choose this mode with
+    intramol=False).
+
+        Parameters
+        ----------
+
+        molecule : BioSimSpace._SireWrappers.Molecule
+            The molecule to be decoupled or annihilated.
+
+        property_map0 : dict
```
lohedges, line 21: The use of `property_map` in this function is inconsistent with how it is used in the rest of the code. It should be treated as a dictionary that allows a user to remap the names of any properties, as desired. For example, a user could change the name of the _element_ property to _kitten_. If we then want to use the search functionality to find specific elements, then we need to pass this mapping to the search function, e.g.:

```python
system.search("element H", property_map={"element" : "kitten"})
```

In this case it appears that you are using the `property_map` as a means of specifying whether the charge or LJ terms are turned on or off in the two end states. If these are the only terms that you need to support, then it would probably be better to add two parameters to the function called `charge` and `LJ` (or `vdw`), which can take a two-element tuple specifying whether these terms are on or off in the end states. Defaults could be `charge=(True, False)`, `LJ=(True, False)`, which would match with what you currently have with the default property maps. (Alternatively you could add four parameters, e.g. `charge0`, `charge1`, `LJ0`, `LJ1`.) 

Also, it might also better to store the decoupling scheme in a different way, i.e. that doesn't conflict with the properties used for the exiting perturbable molecule. Perhaps you could store the decoupling scheme as another property, e.g. something like:

```python
edit_mol = edit_mol.setProperty("decouple", {"charge" : (True, False), "LJ" : (True, False)}).molecule()
```

You can then check to see if the molecule is decoupled by querying the property, then extracting the specific decoupling scheme, e.g.:

```python
if mol._sire_object.hasProperty("decouple"):
    decoupling = mol._sire_object.property("decouple")
    print(decoupling)
    print(decoupling["charge"])
    
Properties(
    LJ => [ 1,0 ],
    charge => [ 1,0 ]
)

[ 1,0 ]
```

test/Sandpit/Exscientia/Free_Energy/test_relative.py
lohedges: Because it is currently not used by `devel`. It will be added when this functionality is moved out of the sandpit. I think it's only just gone on conda-forge, which is great news! (If we want to move there then we can't use any pip packages or vendor any code.)

xiki-tempula:
@lohedges Hi, I wonder if you have reviewed the PR? Then I will start addressing the comments.
Or you are still reviewing it as I didn't see the request changes coming up.

lohedges:
Hi @xiki-tempula. I haven't had a chance to do a full review. I just made specific comments on things that would certainly prevent it from being accepted at present, i.e. those that would break the CI. I also added some comments on the `decouple` function and the use of `property_maps` for passing parameters.

Feel free to address those if you like. I should have more time to look at things tomorrow.

(Ultimately what goes in the sandpit is your (collective) choice, as long as our CI still works. I only make suggestions at this stage to try to keep things consistent with devel where possible. This will help simplify the process of porting things from the sandpit to devel in future.)

xiki-tempula:
@lohedges Thanks. I will address this when the review is finished and keep these things in mind.

xiki-tempula:
@lohedges Hi Lester, I think there is some issue with regard to the current approach.

So the original idea is that 
I will make a self-contained feature in exs/feat_ABC
Then I open a PR to merge exs/feat_ABC to exs/devel.
Then I open a PR to merge exs/devel to BSS/devel

So the current issue is that I have completed a new feature in exs/feat_DEF, I cannot merge this new feature exs/feat_DEF to exs/devel as there is this PR which I try to merge the current exs/devel to BSS/devel.

Given that I'm operating in a tight deadline, do you mind review this PR when I finish the whole ABFE pipeline? Which means that I will push new features into the exs/devel.

lohedges:
Yes, that's no problem. It's probably best if you only PR against our devel when you have fully fledged features anyway. If you'd like some advice/code-review when developing a feature and your end then you can always get in touch, e.g. via slack. You still merge across updates from our devel to yours as needed, then merged these into your own feature branches.

It it's easier, feel free to close this and re-open once the whole ABFE pipeline is ready. That way we won't keep getting CI triggers as you iron out the implementation. (If you do push to `exs/devel` rather than your feature.)

xiki-tempula:
@lohedges Thanks.

PR #331, opened 2022-05-25
Copy __contains__ functionality into Exscientia sandpit.

lohedges:
This PR adds `__contains__` operators to the wrapped Sire objects within the Exscientia sandpit. This allows the user to check whether objects contain other object, e.g.:

```python
# Check whether objects are in a system.
molecule in system
residue in system
atom in system

# Check whether objects are in a molecule.
residue in molecule
atom in molecule

# Check whether an atom is in a residue.
atom in residue
```


msuruzhon (approved):
Looks good @lohedges , many thanks.

Issue #333, opened 2022-05-26
__contain__ cannot handle protein


xiki-tempula:
If one combine protein with ligand, the __contain__ will fail.

```
import BioSimSpace.Sandpit.Exscientia as BSS
>>> ligand = BSS.IO.readMolecules(BSS.IO.glob("test/input/ligands/ligand01*")).getMolecule(0)
>>> protein = BSS.IO.readMolecules(['test/input/molecules/1jr5.crd', 'test/input/molecules/1jr5.top']).getMolecule(0)
>>>print(ligand in (protein+ligand).toSystem()
False
```

lohedges:
This is because it creates a _new_ system, with new and unique molecule numbers, so the original ligand ID doesn't match the one in the new system. You can test whether an object is contained by a system from which it was extracted, not a system that it has become part of.

xiki-tempula:
@lohedges I'm not quite sure if I agree with this logic. But further testing shows that 

`print(ligand in (protein+ligand).toSystem())` gives `False`
`print(ligand in (ligand+protein).toSystem())` gives `True`

lohedges:
Hmm, that's strange. I'll try to see what's gone wrong during the `toSystem()` conversion.

lohedges:
Still don't understand why it doesn't work, but have fixed things by bypassing Sire.

xiki-tempula:
@lohedges I think there might still be some issue with the internal logical
Say we have ligand and protein and combine them to form a system

```
>>> system = (protein+ligand).toSystem()
>>> print(system._sire_object.getMoleculeNumbers())
[MolNum(637), MolNum(636)]
>>> print(protein in system)
True
>>> print(ligand in system)
False
>>> print(ligand._sire_object.number())
MolNum(636)
>>> ligand_new = system[1]
>>> print(ligand_new in system)
True
>>> print(ligand_new._sire_object.number())
MolNum(636)
```

So it seems that even though MolNum is the same you can see different results.

lohedges:
Hmmm, this is very srange. The added unit test check for _exactly_ this behaviour, i.e. creating two systems that combine molecules in opposite orders, then check that both molecules are contained in each system. However, I see the same behaviour for your example, and the molecule numbers are present in the list of `mol_nums`.

I'll debug further and get back to you when I have an update.

xiki-tempula:
@lohedges Sorry the previous message is my fault.
[...]

PR #336, opened 2022-05-26
Fix issue with __contains__ for systems created by adding molecules.

lohedges:
This PR makes the ` __contains__` work as expected when creating new systems by combining molecules. See #333.

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
@@ -189,7 +189,7 @@ class System(_SireWrapper):
                             "'BioSimSpace._SireWrappers.Atom', or 'BioSimSpace._SireWrappers.Residue'.")
 
         # Return whether the object is in the system.
-        return self._sire_object.contains(other._sire_object.molecule())
+        return other._sire_object.molecule().number() in self._mol_nums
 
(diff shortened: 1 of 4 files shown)


python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
```diff
@@ -189,7 +189,7 @@ def __contains__(self, other):
                             "'BioSimSpace._SireWrappers.Atom', or 'BioSimSpace._SireWrappers.Residue'.")
 
         # Return whether the object is in the system.
-        return self._sire_object.contains(other._sire_object.molecule())
+        return other._sire_object.molecule().number() in self._mol_nums
```
msuruzhon, line 192: Are these `mol_nums` unique for each molecule, like `uuid`s? Otherwise, how does `BioSimSpace` keep track of renumbered molecules?

lohedges: Yes, they are unique for each molecule and we additionally enforce the constraint that the numbers are unique when combining molecules to form a new system, i.e. you can't add the same molecule to itself. (Calling `.copy()` on a molecule creates a copy with a new `MolNum`.)

msuruzhon (approved):
Thanks @lohedges this looks good to me.

Issue #338, opened 2022-05-27
Water name is not correctly converted in vacuum simulation


xiki-tempula:
# Prerequisites

This might be an edge case as I'm lazy and find this issue when trying to scavenge some of the existing test.

```
In [1]: import BioSimSpace.Sandpit.Exscientia as BSS

In [2]: system = BSS.IO.readMolecules("test/input/amber/ala/*")

In [3]: m0 = system.getMolecule(0)

In [4]: m1 = system.getMolecule(1)

In [5]: new_system = (m0+m1).toSystem()

In [6]: protocol = BSS.Protocol.Minimisation(steps=100)

In [7]: process = BSS.Process.Gromacs(new_system, protocol, name="test")
/Users/zwu/Documents/GitHub/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py:247: UserWarning: No simulation box found. Assuming gas phase simulation.
  _warnings.warn("No simulation box found. Assuming gas phase simulation.")
/Users/zwu/Documents/GitHub/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/Protocol/_config.py:37: UserWarning: No simulation box found. Assuming gas phase simulation.
  _warnings.warn("No simulation box found. Assuming gas phase simulation.")
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-7-7012fab1aba9> in <module>
----> 1 process = BSS.Process.Gromacs(new_system, protocol, name="test")

~/Documents/GitHub/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py in __init__(self, system, protocol, exe, name, work_dir, seed, extra_options, extra_lines, property_map, restraint, ignore_warnings, show_errors)
    171 
    172         # Now set up the working directory for the process.
--> 173         self._setup()
    174 
    175     def _setup(self):

~/Documents/GitHub/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py in _setup(self)
    224             self.setConfig(self._protocol.getConfig())
    225         else:
--> 226             self._generate_config()
    227         self.writeConfig(self._config_file)
    228 

~/Documents/GitHub/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/Process/_gromacs.py in _generate_config(self)
    758         config = _Protocol.ConfigFactory(self._system, self._protocol)
[...]

PR #339, opened 2022-05-27
Swap water topology before writing custom GRO file for vacuum simulations

lohedges:
This PR addresses issue #338 and ensures that the water topology is swapped to GROMACS format prior to writing the custom Gro87 file needed for vacuum simulations, i.e. where an infinite periodic box is added.


msuruzhon (approved):
Looks good to me, thanks @lohedges .

PR #342, opened 2022-06-14
Add a parameter for Gromacs Process to continue simulation from the previous run. 

kexul:
Hi, This PR adds a new parameter `prev_cpt` for Gromacs Process to continue the simulation from the previous run. 

Context: 
---
I've followed the equilibration protocol (`all restrained short NVT -> backbone restrained NVT -> NVT -> heavy atom restrained  NPT -> NPT`) in [BioSimSpace Tutorial](https://github.com/michellab/BioSimSpaceTutorials/blob/main/04_fep/execution_model/scripts/BSSligprep.py) for a while, using Gromacs as MD engine, instead of  PMEMD.  But my simulation always suffered from `#WARNING SUBSAMPLING ENERGIES RESULTED IN LESS THAN 50 SAMPLES` when analyzing the result. 

As Julien suggested [here](https://github.com/michellab/Sire/issues/90#issuecomment-165442941), it means that the simulation does not start from a well-equilibrated trajectory (I've checked the overlap matrix, which seems to be reasonable). I checked my simulation and found that in the `all restrained short NVT -> backbone restrained NVT -> NVT -> heavy atom restrained  NPT -> NPT` pipeline, the temperature always starts from 0 in each step. 

In Gromacs [tutorial](http://www.mdtutorials.com/gmx/lysozyme/07_equil2.html), the NPT equilibration should be started from a thermalized system, so I added this parameter and continued the simulation from the previous step, now the subsampling warning disappeared.

### DIFF
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -131,11 +131,11 @@ class Gromacs(_process.Process):
                                             "Please install GROMACS (http://www.gromacs.org).")
 
         if not isinstance(ignore_warnings, bool):
-            raise ValueError("'ignore_warnings' must be of type 'bool.")
+            raise ValueError("'ignore_warnings' must be of type 'bool'.")
         self._ignore_warnings = ignore_warnings
 
         if not isinstance(show_errors, bool):
-            raise ValueError("'show_errors' must be of type 'bool.")
+            raise ValueError("'show_errors' must be of type 'bool'.")
         self._show_errors = show_errors
 
         # Initialise the stdout dictionary and title header.
@@ -160,8 +160,16 @@ class Gromacs(_process.Process):
         # Initialise the PLUMED interface object.
         self._plumed = None
 		
-		# Set the path of Gromacs checkpoint file
-        self._prev_cpt = prev_cpt
+		# Set the path of Gromacs checkpoint file.
+        self._prev_cpt = None
+        if prev_cpt is not None:
+            if not isinstance(prev_cpt, str):
+                raise ValueError("'prev_cpt' must be of type 'str'.")
+            else:
+                if _os.path.isfile(prev_cpt):
+                    self._prev_cpt = prev_cpt
+                else:
+                    raise IOError("Gromacs checkpoint file doesn't exist: '%s'" % prev_cpt)
 
         # Now set up the working directory for the process.
         self._setup()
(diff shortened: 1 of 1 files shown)


lohedges:
Thanks for this. We currently only have a `restart` option for the `Production` protocol. I think I'd prefer to extend this to all protocols that could support it, possibly moving into the `Process` as you have done. Part of the problem is the way that restarts are handled by the different engines, and whether it is possible to pick up more complex state from a partially completed protocol, such as when the temperature is varied.

I'll have a think about this when I get a chance and get back to you with suggestions.

python/BioSimSpace/Process/_gromacs.py
```diff
@@ -156,6 +159,9 @@ def __init__(self, system, protocol, exe=None, name="gromacs",
 
         # Initialise the PLUMED interface object.
         self._plumed = None
+		
+		# Set the path of Gromacs checkpoint file
+        self._prev_cpt = prev_cpt
```
lohedges, line 164: There is no type or value checking going on here. What should `prev_cpt` be (presumably a `str` type) and what happens when it's not?

python/BioSimSpace/Process/_gromacs.py
```diff
@@ -314,6 +320,8 @@ def _generate_config(self):
             config.append("nstlog = %d" % report_interval)      # Interval between writing to the log file.
             config.append("nstenergy = %d" % report_interval)   # Interval between writing to the energy file.
             config.append("nstxout = %d" % restart_interval)    # Interval between writing to the trajectory file.
+            if self._prev_cpt is not None:
```
lohedges, line 323: This would be `True` for anything other than a `NoneType`, i.e. the use could pass `False`, `'hello'`, etc.

python/BioSimSpace/Process/_gromacs.py
kexul: Thanks for the feedback @lohedges , I've added a type check and make sure it's a valid file.

python/BioSimSpace/Process/_gromacs.py
kexul: Now `self._prev_cpt` could be `NoneType` or valid file path.

python/BioSimSpace/Process/_gromacs.py
```diff
@@ -820,7 +828,12 @@ def _generate_binary_run_file(self):
                   "/%s.out.mdp" % _os.path.basename(self._config_file).split(".")[0]
 
         # Use grompp to generate the portable binary run input file.
-        command = "%s grompp -f %s -po %s -c %s -p %s -r %s -o %s" \
+        if self._prev_cpt is not None:
```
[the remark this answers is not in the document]
kexul: Now, self._prev_cpt could be NoneType or a valid file path.

lohedges:
Just to note that I've made a few minor formatting tweaks on `devel` and changed the option name to `checkpoint_file` for clarity.

PR #345, opened 2022-06-22
Apply GROMACS water topology swap for all protocols.

lohedges:
This PR applies some more fixes for #338. Previously the fix was only applied to specific protocols. Now we catch them all. (I'm not sure why the CI didn't catch this previously.)

Thanks @xiki-tempula for catching this!


msuruzhon (approved):
Thanks @lohedges this looks good to me.

PR #347, opened 2022-07-01
Use system with unique numbering to determine perturbable residue number

lohedges:
This PR ensures that the `self._renumbered_system` attribute is used to get the residue number of the perturbable molecule, which is required as a configuration file option. Previously, the _original_ system was used, so it may have not worked as expected if there were duplicate residue numbers in the system, which is likely when creating protein ligand complexes for RBFE simulations.

I have also ported the functionality across to the Exscientia Sandpit, where the `perturbed residue number` option was missing from `ConfigFactory.generateSomdConfig`.

Finally, I have also a minimal test to validate that the config entry  is correct when the perturbable molcule is placed at different positions within a system.

This PR also raised a point regarding the use of pickled objects for unit tests. These will need to be created for _all_ Sandpits where the test exists, so that `isinstance` checks will work, i.e. a pickled System from the main code won't work as a System object in `BioSimSpace.Sandpit.Exscientia`, and vice-versa. (This is the correct behaviour, since they are _different_ types, which could diverge over time if new attributes are added.


python/BioSimSpace/Process/_somd.py
```diff
@@ -577,7 +578,8 @@ def _generate_config(self):
             self.addToConfig("lambda_val = %s" \
                 % self._protocol.getLambda())                                       # The value of lambda.
 
-            res_num = self._system.search("resname LIG")[0]._sire_object.number().value()
+            res_num = _System(self._renumbered_system) \
+                .search("resname LIG")[0]._sire_object.number().value()
```
msuruzhon, line 582: Is it better to detect it by searching for perturbable residues instead and asserting that there is only one? Relying on resname is probably a bit unreliable.

lohedges: For SOMD, it requires a specific name, i.e. `LIG` in this case, which is enforced elsewhere during the merging. I agree that it's probably not super reliable depending on how the system was generated, though. I'll change the search to make it more robust.

Cheers.

lohedges: Oh, and we also enforce that there only a single perturbable molecule here:

https://github.com/michellab/BioSimSpace/blob/950c35a95d08ad0d3e7c6fd310e4d49dd61def9e/python/BioSimSpace/Process/_somd.py#L225-L226

This is before the `res_num` is computed.

msuruzhon (approved):
Thanks @lohedges this looks good to me now.

Issue #340, opened 2022-05-30
Improve interoperablilty of MDTraj/MDAnalysis trajectory handling


xiki-tempula:
Say if I create a BSS.Trajectory.Trajectory object with

```
traj = Trajectory(None,
    "test/Sandpit/Exscientia/input/protein_ligand/traj.xtc",
    "test/Sandpit/Exscientia/input/protein_ligand/complex.tpr")
```
Then I cannot call traj.getFrames() as traj._trajectory is None

xiki-tempula:
Ok, I see the issue, by default, BSS.Trajectory.Trajectory will use mdtraj to read the topology and thus, it will fail to read the Gromacs tpr file as topology, which is only supported by MDAnalysis.

lohedges:
Yes, that's right. I'll need to add a check for a `tpr` topology extension and use MDAnalysis as the backend in that situation. (This is handled automatically when using a `Process`, since it knows what the engine is.)

lohedges:
We've also had situations where some of our users don't want MDAnalysis in their environment since it conflicts with other packages, hence defaulting to MDTraj.

xiki-tempula:
@lohedges Thanks, the whole getFrames() method might also need dome polishing if one assumes that the traj._trajectory could be an MDA object.

lohedges:
I'll add some functionality that tries to load the files using either backend, then flag internally which one worked. Then we can simply use the same engine when extracting frames, etc.

lohedges:
Closing for now as this has been addressed in #353. Feel free to open another thread if other trajectory handling issues arise.

Cheers.

PR #353, opened 2022-07-05
Improve interoperablilty of MDTraj/MDAnalysis trajectory handling

lohedges:
This PR addresses #340 and improves the interoperability of trajectory handling within BioSimSpace. You can now use the `format="auto"` option to attempt to return a trajectory object via any supported backend, i.e. this would try MDTraj, then MDAnalysis if that failed.

The rudimentary RMSD analysis now supports both backends and unit tests validate that the extracted values are consistent in both bases.

Note that the sub-package is still fairly bare-bones. It is largely used internally, i.e. when binding a `Trajectory` object to a process. We mostly expect users to simply return a trajectory object in their chosen format and work with it directly. This PR streamlines that process and makes it more reliable.


python/BioSimSpace/Sandpit/Exscientia/Trajectory/_trajectory.py
```diff
@@ -202,9 +208,24 @@ def getTrajectory(self, format="mdtraj"):
...
         """
 
-        if format.upper() not in ["MDTRAJ", "MDANALYSIS"]:
+        # Strip whitespace and convert to upper case.
+        format = format.replace(" ", "").upper()
+
+        if format.replace(" ", "").upper() not in ["MDTRAJ", "MDANALYSIS", "AUTO"]:
             _warnings.warn("Invalid trajectory format. Using default (mdtraj).")
-            format = "mdtraj"
+            format = "MDTRAJ"
```
xiki-tempula, line 216: Would it be best to default to `AUTO`

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
```diff
@@ -0,0 +1,48 @@
+import BioSimSpace.Sandpit.Exscientia as BSS
+
+import pytest
+
+@pytest.fixture
+def traj_mdtraj(scope="session"):
+    """A trajectory object using the MDTraj backend."""
+    return BSS.Trajectory.Trajectory(
+            trajectory="test/Sandpit/Exscientia/input/trajectories/ala.trr",
+            topology="test/Sandpit/Exscientia/input/trajectories/ala.tpr")
```
xiki-tempula, line 10: I think you mentioned that mdtraj cannot handle tpr as topology file?

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
```diff
@@ -0,0 +1,48 @@
...
+
```
xiki-tempula, line 11: Do you mind add a frame to test `traj.getFrames(index)`? Thanks.

python/BioSimSpace/Sandpit/Exscientia/Trajectory/_trajectory.py
lohedges: Yes, it could be, I was just being consistent with the existing behaviour so as not to surprise anyone. I can update and check that it doesn't break anything.

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
lohedges: Yes, can do. It's implicitly tested in the coordinates check, but I can make it more explicit.

python/BioSimSpace/Sandpit/Exscientia/Trajectory/_trajectory.py
xiki-tempula: @lohedges So currently, if my understanding is correct, when backend is set to auto, it will try to use MDTRAJ first. So unless the mdtraj failed, I would imagine that it would produce the same result?

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
xiki-tempula: @lohedges Thanks. I wonder what do you mean `naming was irrelevant`. I would imagine that if you try to feed `tpr` with backend set to `mdtraj`, it should raise an error and say that it cannot read 'tpr' file?

lohedges: Yes, but now that we have the "auto" option it would just fallback to using `MDAnalysis`. In the tests, the naming was just to infer that I was passing files which would only work with a given backend, i.e. the named one, so I knew that I was testing two different backends, rather than the same one. I had effectively named them the round way around, which had no impact on the tests.

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
lohedges: Just to check, do you mean the standalone `Trajectory.getFrame` function? This extracts a frame from the trajectory without loading the whole thing into memory, so is handy for getting specific frames from large files. Currently this only uses `mdtraj`. I can try to adapt it to use both backends, but I'm not sure if that's possible.

lohedges: Yes, it works with both backends, although it's a bit slower with `MDAnalysis`. Will try to re-use the existing functionality so that `getFrame` behaves the same way as the `getTrajectory` method, only extracting a single frame directly.

xiki-tempula: Thanks. My user case is
```
traj = Trajectory(process=None, trajectory='gromacs.xtc', topology='gromacs.tpr')
system = traj.getFrames(0)
```

lohedges: This works for me with the current code. At least with the example input used for the tests.

lohedges:
```python
In [15]: config = process.getConfig()

In [16]: config.append("nstxout-compressed = 10000")

In [17]: process.setConfig(config)

In [18]: process.start()
Out[18]: BioSimSpace.Process.Gromacs(<BioSimSpace.System: nMolecules=631>, <BioSimSpace.Protocol.Custom>, exe='/usr/local/gromacs/bin/gmx', name='gromacs', work_dir='/home/lester/Code/BioSimSpace/demo/tmp', seed=None)

In [19]: process.wait()

In [20]: traj = BSS.Trajectory.Trajectory(trajectory="tmp/gromacs.xtc", topology="tmp/gromacs.tpr")

In [21]: traj
Out[21]: <BioSimSpace.Trajectory: nFrames=11, backend='mdanalysis'>

In [22]: traj.getFrames(0)
Out[22]: [<BioSimSpace.System: nMolecules=1>]
```

xiki-tempula (approved):
LGTM

test/Sandpit/Exscientia/Trajectory/test_trajectory.py
lohedges: One issues with `getFrames` is that, depending on the format, we can lose the original molecules in the system, i.e. what is returned is just a single molecule. Process objects have a way of remapping things, or using alternative means to extract frames, e.g. `trjconv`. Here we are just wrapping around `MDTraj` and `MDAnalysis`, so are limited by what they can read/write.

xiki-tempula: So if a process is being passed to the trajectory, would the property of the system in the process be retained? Or it will be lost anyway?

xiki-tempula: Thanks for the explanation. I'm happy with this PR being merged. @msuruzhon asked me to review this and I guess I will test myself to see if it works for my purpose.

PR #355, opened 2022-07-08
Allow user to specify whether to use connectivity or coordinates during HMR.

lohedges:
This PR adds an additional option, `use_coordinates`, that allows the user to specify whether to use the current molecular coordinates to compute the connectivity during hydrogen mass repartitioning. If `False`, the `connectivity` property will be used, i.e. the bonding from the topology generated by the force field.

The PR also updates the unit tests to ensure that the repartitioning is invertible.

Edits are backwards compatible, but will need the latest Sire (currently building) to ensure that the `connectivity` is used by default.

Cheers.

### DIFF
--- a/python/BioSimSpace/Process/_gromacs.py
+++ b/python/BioSimSpace/Process/_gromacs.py
@@ -822,7 +822,6 @@ class Gromacs(_process.Process):
 
         # Add the default arguments.
         self.setArg("mdrun", True)          # Use mdrun.
-        self.setArg("-v", True)             # Verbose output.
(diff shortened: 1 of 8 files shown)


msuruzhon (approved):
This is great, many thanks for that @lohedges.

lohedges:
Thanks. I've just pushed a minor update to also disable verbose output for GROMACS by default, as requested by @xiki-tempula. (This was needed for getting the estimated end time interactively, which was never actually implemented.)

I'll merge once the Sire build has finished.

PR #357, opened 2022-07-12
Feature smd

AdeleHardie:
Patches to running more complex sMD simulations using RMSD CV(s)

PR #358, opened 2022-07-12
Port recent PRs to devel into Exscientia sandpit

lohedges:
This PR ports changes from two recent PRs, #342 and #357 into the Exscientia sandpit. The first introduces a `checkpoint_file` option to `Process.Gromacs` to enable restarting of `Equilibration` protocols. (This is a WIP, hence why it doesn't work for all protocols at present. We want to find a way for this to work interoperably across protocols and engines, which is tricky.) The second fixes some issues with the steered molecular dynamics implementation.


lohedges:
I think that's just a CI glitch. It should be resolved by a re-run. Cheers.

PR #363, opened 2022-07-15
Add the ABFE functionality

xiki-tempula:
This PR syncs the EXS/dev to this dev and the major update is the ABFE functionality.

Type of change
Bug fix (non-breaking change which fixes an issue)

- Fix the unit test to have them running fine on GMX 2022.1
- Change the maxwarn to use the Gromacs unlimited option
- Correct some formatting issue with the checkpoint_file in the doc string
- Allow the BSS.FreeEnergy.Relative to recognise the conda-forge gromacs
- Allow BSS.FreeEnergy.Relative to load the temperature as float
- The velocity is generated automatically for BSS.Protocol.Equilibration

New feature (non-breaking change which adds functionality)

- Add the Github actions such that the CI will run all the unit tests
- Use pd.DataFrame and pd.Series internally in the BSS.Protocol._FreeEnergyMixin such that the more than one lambda state is supported.
- Add setLambda method to BSS.protocol.free_energy_mixin
- When supplying a Protocol with multiple lambda state, the BSS.Protocol._config could write multiple stage lambda into Gromacs mdp files
- Change the default mdp options to better suit the gmx-2022
- Allow BSS.FreeEnergy.relative to use the new lambda config
- Write test to ensure that the mdp file has been modified
- Allow the BSS.FreeEnergy.Relative to handle decoupled molecule
- Allow the BSS.Protocol.Config to edit the mdp file such that decoupled molecule could be written
- Add new method BSS.Align.Decouple to make molecule as being decoupled
- Allow BSS.mol and BSS.system to handle decoupled state properly
- Add the new BSS.FreeEnergy.RestraintSearch class to deal with ABFE restraint search.
- Allow the BSS.FreeEnergy.Relative to handle restraint
- Allow the BSS.FreeEnergy.Relative to use pd.DataFrame internally to handle the lambda values.
- Remove the duplication of the config related lines in BSS.Process.Gromacs
- Change the default `refcoord-scaling` to `com`
- Allow BSS.Process.Gromacs to handle restraint
- Add new tests for the ligand-protein system

Changes (breaking change which adds functionality)

- Change the folder being created to use the index of the lambda (e.g. lambda_0, lambda_1 ... lambda_32)
- Change the default trajectory saving format to xtc from trr.
- Change the mdp options such that they are GPU compatible


lohedges:
Thanks for this, I should be able to merge in within the next few days.

lohedges:
Hi there, yes, I meant review as well as merge. Since it's quite a large PR it might take me a while to go through. I'll let you know if I have any comments/questions.

Cheers.

lohedges:
I've now done the work required to update your PR so that it is compliant with the new Sire 2023 API. I've checked that all tests pass locally and have pushed my edits to the [merge-exs-devel](https://github.com/michellab/BioSimSpace/tree/merge-exs-devel) branch.

Do you want to fetch this locally and check that everything works as expected?

With regards to this PR, I'm not sure if it's best that I submit a PR of my branch against your devel, then you re-open this PR with the updated code? Alternatively, I could still go through the code here and submit a review. Any edits/changes could then be incorporated in the the updated branch.

Cheers.

lohedges:
Oh, and just to note that Sire 2023 has [broken SOMD](https://github.com/michellab/Sire/issues/408). I have a [fix](https://github.com/michellab/Sire/tree/fix-somd-2023) and this will be merged in as soon as it has been tested.

xiki-tempula:
@lohedges Thanks for the help. Is solving the conflict very difficult?
I think the best route will be create a branch from Exscientia:devel to be Exscientia:michellab, then merge michellab:devel to Exscientia:michellab, then create a PR to merge Exscientia:michellab to Exscientia:devel.
I will have a test of merge-exs-devel now, is this branch using Sire 2023?

lohedges:
No the conflict resolution wasn't too bad. It was just accepting a few updates your branch over `devel`, renaming all `Sire imports` to `from sire.legacy` instead, and deleting the, now redundant, `fast_system.py` file.

Yes, the branch I have made requires Sire 2023 to function. I think I forgot to update this in your CI/build scripts, so those will likely fail. (I just ran the tests locally.)

Assuming that you are okay with the branch, we could always keep this PR, then I just re-play my edits on top following the merge. I have a few review comments that I need to add first, but these mostly concern formatting, rather than the actual ABFE functionality.

xiki-tempula:
Waiting for https://github.com/michellab/Sire/issues/407 to be fixed.

xiki-tempula:
I have merged the dev to our dev. I wonder if you mind review this PR when you are free, please? Thank you.

.github/workflows/Sandpit_exs.yml
```diff
@@ -0,0 +1,53 @@
...
+          miniforge-variant: Mambaforge
+          use-mamba: true
+
+      - name: Install dependency
+        run: |
+          # gromacs=2022.2 as gromacs=2022.3 is not available on OSX ATM
+          mamba install -c conda-forge -c michellab/label/dev biosimspace python=3.8 ambertools gromacs=2022.2 "sire>=2023.0.0" alchemlyb pytest
+          python -m pip install git+https://github.com/Exscientia/MDRestraintsGenerator.git 
+          # For the testing of BSS.FreeEnergy.Relative.analysis
+          python -m pip install https://github.com/alchemistry/alchemtest/archive/master.zip
```
lohedges, line 42: Would it make sense to install `alchemlyb` via conda, since it is now available in conda-forge? This will be the version that we will ultimately depend on, so would be good to be consistent, e.g. if the versions happen to be different, or the conda environment can only be resolved with a particular version.

python/BioSimSpace/Sandpit/Exscientia/Align/_decouple.py
```diff
@@ -0,0 +1,91 @@
...
+    # Validate input.
+
+    if not isinstance(molecule, _Molecule):
+        raise TypeError("'molecule' must be of type 'BioSimSpace._SireWrappers.Molecule'")
+
+    # Cannot decouple a perturbable molecule.
+    if molecule.isDecoupled():
+        raise _IncompatibleError("'molecule' has already been decoupled!")
+
+    for field, name in zip((charge, LJ), ('charge', 'LJ')):
```
lohedges, line 55: This is assuming that `charge` and `LJ` are iterable. It would probably be best to first check that they are both either a `list` or `tuple`, then do this check, i.e. put it inside:

```python
if isinstance(charge, (list, tuple)) and ininstance(LJ, (list, tuple)):
    ...
else:
    raise TypeError(...)
```

(You might want to do this independently for each parameter, so you can specify which one is wrong.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -44,6 +44,7 @@
 
 try:
```
lohedges: It looks like `alchemlyb` is a `noarch` conda package, e.g. see [here](https://anaconda.org/conda-forge/alchemlyb). As such, it's availability on given platforms is governed by the availability of its dependencies. Could this `try` except block be replaced with the `_try_import` and `_have_imported` functions from `BioSimSpace._Utils`. These allow us to handle packages that aren't present on all architectures. For example, see how we deal with RDKit [here](https://github.com/michellab/BioSimSpace/blob/303ea91f998da1c64e67734f4f970f2e3d48f55d/python/BioSimSpace/Align/_align.py#L56-L61).

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -217,7 +223,7 @@ def __init__(self, system, protocol=None, work_dir=None, engine=None,
                     raise _MissingSoftwareError("Cannot use GROMACS engine as GROMACS is not installed!")
 
                 # The system must have a perturbable molecule.
-                if system.nPerturbableMolecules() == 0:
+                if system.nPerturbableMolecules() == 0 and system.nDecoupledMolecules() == 0:
                     raise ValueError("The system must contain a perturbable molecule! "
```
lohedges: Should this now say "a perturbable or decoupled molecule"?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -44,6 +44,7 @@
 
 try:
     import alchemlyb as _alchemlyb
+    from alchemlyb.workflows import ABFE
     from alchemlyb.postprocessors.units import R_kJmol, kJ2kcal
```
lohedges: These should be imported privately, e.g. as `_R_kJmol` and `_kJ2kcal`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -656,193 +564,46 @@ def _analyse_gromacs(work_dir=None, estimator=None):
...
-                                found_temperature = True
-                                break
-
-                if not found_temperature:
-                    raise ValueError(f"The temperature was not detected in the GROMACS output file, {file}")
-
-            if temperatures[0] != temperatures[-1]:
-                raise ValueError("The temperatures at the endstates don't match!")
-
+        if is_alchemlyb:
```
lohedges: If you use the new `_try_import` functionality, then this will fail if the package wasn't imported, i.e. we only fail when functionality is _used_, not when it's _imported_.

(I note that this PR is actually also commenting on @annamherz's edits, so it's a bit confusing.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -0,0 +1,343 @@
...
+from sire.legacy.Units import mole as Sire_mole
+
+from .._SireWrappers import Atom
+from ..Types import Length, Angle, Temperature
+from .._SireWrappers import System as _System
+from ..Units.Length import nanometer
+from ..Units.Area import nanometer2
+from ..Units.Temperature import kelvin
+from ..Units.Angle import degree, radian
+from ..Units.Energy import kj_per_mol, kcal_per_mol
```
lohedges: Can all of these be imported privately.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -0,0 +1,343 @@
...
+                              "phiC0": BioSimSpace.Types.Angle},
+        "force_constants":{"kr": BioSimSpace.Types.Energy / BioSimSpace.Types.Area,
+                           "kthetaA": BioSimSpace.Types.Energy / (BioSimSpace.Types.Area * BioSimSpace.Types.Area),
+                           "kthetaB": BioSimSpace.Types.Energy / (BioSimSpace.Types.Area * BioSimSpace.Types.Area),
+                           "kphiA": BioSimSpace.Types.Energy / (BioSimSpace.Types.Area * BioSimSpace.Types.Area),
+                           "kphiB": BioSimSpace.Types.Energy / (BioSimSpace.Types.Area * BioSimSpace.Types.Area),
+                           "kphiC": BioSimSpace.Types.Energy / (BioSimSpace.Types.Area * BioSimSpace.Types.Area)}}
+
+    '''
+    def __init__(self, system, restraint_dict, temperature, rest_type='Boresch'):
```
lohedges: I'd prefer not to mix formatting in parameter names, e.g. `restraint_dict` and `rest_type`. Can this be `restraint_dict` and `restraint_type`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -0,0 +1,343 @@
...
+    atom1-atom2-... (equilibrium value, force constant)
+
+    The nomenclature:
+    Bonds: r1-l1 (r0, kr)
+    Angles: r2-r1-l1 (thetaA0, kthetaA), r1-l1-l2 (thetaB0, kthetaB)
+    Dihedrals: r3-r2-r1-l1 (phiA0, kphiA), r2-r1-l1-l2 (phiB0, kphiB), r1-l1-l2-l3 (phiC0, kphiC)
+
+    The restraint_dict has the following compact format.
+
+    restraint_dict = {
```
lohedges, line 62: Is there a reason to pass all of the restraint information as a dictionary, rather than separate parameters? Perhaps it makes it easier to pass around elsewhere. If so, it could always be packed into a a dictionary internally.

Ultimately we might be able to do this using `Bond`, `Angle`, and `Dihedral` objects when these are exposed via BioSimSpace. (We have `Bond` already, and I imagine the others might be available by the full 2023 release.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -0,0 +1,343 @@
...
+           Parameters
+           ----------
+
+           engine : str
+               The molecular dynamics engine used to generate the restraint.
+               Available options currently is "GROMACS" only. If this argument
+               is omitted then BioSimSpace will choose an appropriate engine
+               for you.
+        """
+        if engine.lower() == 'gromacs':
```
lohedges: You might also want to strip white-space to make it more robust, e.g. `engine.replace(" ", "").lower()`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+    from alchemlyb.parsing.amber import extract_u_nk as _amber_extract_u_nk
+    from alchemlyb.parsing.amber import extract_dHdl as _amber_extract_dHdl
+    from alchemlyb.preprocessing.subsampling import statistical_inefficiency as _statistical_inefficiency
+    from alchemlyb.estimators import AutoMBAR as _AutoMBAR
+    from alchemlyb.estimators import TI as _TI
+    from alchemlyb.postprocessors.units import to_kcalmol as _to_kcalmol
+    is_alchemlyb = True
+except:
+    print('Please install alchemlyb via pip for analysis using it.')
+    is_alchemlyb = False
```
lohedges: Use the `_try_import` and `_have_imported` functions.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+
+try:
+    import MDRestraintsGenerator
+    from MDRestraintsGenerator import search as _search
+    from MDRestraintsGenerator.restraints import FindBoreschRestraint as _FindBoreschRestraint
+    is_MDRestraintsGenerator = True
+except:
+    print('Please install MDRestraintsGenerator for analysis using it.')
+    is_MDRestraintsGenerator = False
```
lohedges: Use the `_try_import` function. You can specify an `install_command` to tell the user how to get the package. See [here](https://github.com/michellab/BioSimSpace/blob/303ea91f998da1c64e67734f4f970f2e3d48f55d/python/BioSimSpace/_Utils/_module_stub.py#L87) for details.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+from ..Units.Angle import radian as _radian
+from ..Units.Angle import degree as _degree
+from ..Units.Energy import kcal_per_mol as _kcal_per_mol
+
+from ..MD._md import _find_md_engines
+
+if _is_notebook:
+    from IPython.display import FileLink as _FileLink
+
+from ._restraint import Restraint
```
lohedges: Make this private.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+        if protocol is not None:
+            if isinstance(protocol, _Protocol.Production):
+                self._protocol = protocol
+            else:
+                raise TypeError("'protocol' must be of type 'BioSimSpace.Protocol.Production'")
+        else:
+            self._protocol = _Protocol.Production()
+
+        self._extra_options = extra_options if extra_options is not None else {}
+        self._extra_lines = extra_lines if extra_lines is not None else []
```
lohedges: I think this is unrelated to this PR, but what happens if these aren't lists or dictionaries? There is no type checking here.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+                work_dir=self._work_dir, extra_options=self._extra_options,
+                extra_lines=self._extra_lines, **kwargs)
+
+
+    @staticmethod
+    def analyse(work_dir, system, traj, temperature, rest_type='Boresch',
+                method='MDRestraintsGenerator',
+                append_to_lig_selection="",
+                recept_selection_str='protein and name CA C N',
+                cutoff=10): # In Angstrom
```
lohedges: Could use a `BioSimSpace.Types.Length` object to allow unit conversion.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+               Default is 'Boresch'.
+
+           method: str
+               The method to use to derive the restraints. Currently only 'MDRestraintsGenerator'
+               is supported.
+
+           append_to_lig_selection: str
+               Appends the supplied string to the default atom selection which chooses
+               the atoms in the ligand to consider as potential anchor points. The default
+               atom selection is f'resname {ligand_resname} and not name H*'. Uses the
```
lohedges, line 304: These format strings aren't expanded in the docstrings, e.g. if you type:

```python

help(BSS.FreeEnergy._restraint_search.RestraintSearch._analyse)
```

Gives:

```
...
append_to_lig_selection: str
           Appends the supplied string to the default atom selection which chooses
           the atoms in the ligand to consider as potential anchor points. The default
           atom selection is f'resname {ligand_resname} and not name H*'. Uses the
           mdanalysis atom selection language. For example, 'not name O*' will result
           in an atom selection of f'resname {ligand_resname} and not name H* and not
           name O*'.
...
```

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+               mdanalysis atom selection language. For example, 'not name O*' will result
+               in an atom selection of f'resname {ligand_resname} and not name H* and not
+               name O*'.
+
+           recept_selection_str: str
+               The selection string for the atoms in the receptor to consider
+               as potential anchor points. The default atom selection is
+               'protein and name CA C N'. Uses the mdanalysis atom selection
+               language.
```
lohedges: I guess `MDRestraintsGenerator` relies on `MDAanalysis` so there is no way of doing this search natively. As long as we explain what this is and the documentation is easy enough to find, we should be okay.

Is it possible to make the parameter naming a little more consistent. What gets abbreviated is a little random, e.g. here we have `recept` and `str`, but keep all of `selection`. I'm fine with `str`, since it's a buitin type, but might be good to be verbose with `recept` and `lig`, or explain what they are somewhere. (Obviously someone using this will know, but good to be able to understand from the docs and function naming.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+                u,
+                l_atoms[0],
+                p_selection=recept_selection_str)
+            psearch.run()
+            atom_set.extend(
+                [(l_atoms, p) for p in psearch.host_atoms])
+        # Create the boresch finder analysis object
+        boresch = _FindBoreschRestraint(u, atom_set)
+        # Run the restraint analysis
+        boresch.run()
```
lohedges: What happens when things go wrong here, i.e. what does `MDRestraintsGenerator` do? I prefer to wrap call to external libraries in a try / except block so we can give a short-form BioSimSpace exception, followed by the longer message from the tool. This is helpful when errors are extremely verbose, or cryptic. It is also helpful from a users perspective since they are calling a BioSimSpace function, so don't know anything about the backend used to do the calculation.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+                                "kthetaB": kthetaB,
+                                "kphiA": kphiA,
+                                "kphiB": kphiB,
+                                "kphiC": kphiC
+                                }}
+        # TODO: extract the best frame and feed it into Restraint
+        # Waiting for the BSS to fix the getFrames
+        # best_frame = traj.getFrames(index)
+        best_frame = system
+        restraint = Restraint(best_frame, restraint_dict,
```
lohedges: Looking here, it seems like a restraint is something that is generated _for_ the user. If so, perhaps we want to make `Restraint` private if it is only used internally. Are we expecting to support user-defined restraints, or will they always get one as the result of a `RestraintSearch`.

I'm only thinking about this as a way of minimising the public API and making things robust, i.e. not allowing the user to create objects that they don't need to, such as an invalid restraint. In this case your `restraint_dict` makes more sense, since it's not something that the user should ever generate themselves.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
lohedges: Just to note that Sire 2023 will also have trajectory handling (not sure what formats, yet) and will be able to search/return properties from different frames. This might make it easier to provide this functionality natively. (Definitely out of scope for now.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+            axs[i].axvline(x=boresch_dof_dict[pair][dof]["avg"], color='r',
+                           linestyle='dashed', linewidth=2, label="mean")
+            if dof == "r":
+                axs[i].set_xlabel("r ($\AA$)")
+            else:
+                axs[i].set_xlabel(f"{dof} (rad)")
+            axs[i].set_ylabel("Num Vals")
+            axs[i].legend()
+        fig.tight_layout()
+        fig.savefig(f'{work_dir}/boresch_dof_hist.png', facecolor="white")
```
lohedges, line 1032: Does plotting always work? What happens if the user is logged in on a server without an X session?

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
```diff
@@ -1,5 +1,10 @@
 __all__ = ["_FreeEnergyMixin"]
 
+import warnings
+
+import numpy as np
+import pandas as pd
```
lohedges: Import privately.

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
```diff
@@ -18,16 +23,16 @@ def __init__(self,
...
+           lam_vals : [float] or pandas.DataFrame
+               A list of lambda values.
 
-           min_lam : float
+           min_lam : float or pandas.Series
                The minimum lambda value.
 
-           max_lam : float
+           max_lam : float or pandas.Series
                The maximum lambda value.
```
lohedges, line 36: I was just wondering if there was any need to use `pandas.Series` and `pandas.DataFrame`, i.e. could these just be Python dictionaries? In your tests you just pass dictionaries to the constructors, e.g.:

```python
            lam_vals=pd.DataFrame(
                data={'bonded': [0.0, 0.25, 0.5, 0.75, 1.0, 1.0, 1.0],
                      'coul': [0.0, 0.0, 0.0, 0.5, 1.0, 1.0, 1.0],
                      'vdw': [0.0, 0.0, 0.0, 0.0, 0.0, 0.5, 1.0]}))
```

This seems like an additional layer of abstraction that's not needed. (The user might be familiar with Python objects, but not `pandas`.) We currently implicitly depend on `pandas` via other dependencies, so would need to add a hard dependency if used.

If it's convenient to use `pandas` internally, i.e. it makes the validation simpler, e.g. checking all entries have the same number of values, etc., then the data could be stored that way behind the scenes.

(Ultimately I don't mind what you use in the Sandpit, since it is what is convenient for you. Perhaps you are generating these inputs elsewhere, so `Series` and `DataFrame` make sense.)

python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
```diff
@@ -812,6 +812,34 @@ def nPerturbableMolecules(self):
...
+           molecules : [:class:`Molecule <BioSimSpace._SireWrappers.Molecule>`]
+               A list of decoupled molecules.
+        """
+
+        molecules = []
+
+        for mol in self:
+            if mol.isDecoupled():
+                molecules.append(_Molecule(mol))
+        return molecules
```
lohedges, line 830: WIth Sire 2023 you _should_ be able to avoid the loop by searching for the property directly, e.g.:

```python
molecules = []
for mol in self.search("property decouple").molecules():
    molecules.append(mol)
```

This will be _much_ faster for large systems.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
```diff
@@ -656,193 +564,46 @@ def _analyse_gromacs(work_dir=None, estimator=None):
...
-            with _tempfile.TemporaryDirectory() as tmpdirname:
-                file = f'{tmpdirname}/version.txt'
-                with open (file,'w') as f:
-                    print(p, file=f)
-                f = open(file,'r')
-                for line in f:
-                    if "GROMACS version" in line:
-                        l = line.strip().split(':')
-                        l = l[1].strip()
-                        l = float(l)
```
lohedges, line 671: I appreciate that this is likely not your edit, but can we not use a single character `l` as a variable name. Depending on the font, this will look identical to `1`. Here we could just safely re-use `line`. (Although it seems a bit weird for the same variable to refer to multiple different things, i.e. the full line, the split line as a list, the first record with no preceding/trailing whitespace, the first record as a float. Perhaps `tmp` would be best to identify that it's a temporary variable.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
```diff
@@ -0,0 +1,1113 @@
...
+
+       Returns
+       -------
+
+       restraint : :class:`Restraint <BioSimSpace.Sandpit.Exscientia.FreeEnergy.Restraint>`
+           The restraints of `rest_type` which best mimic the strongest receptor-ligand
+           interactions.
+        """
+        # Please don't review this part yet as Finlay is working on it.
+        # TODO Tidy this up and improve. Change algorithm to fit to match that discussed.
```
lohedges, line 770: Sorry, didn't see this bit. Will discuss with Finlay separately. Perhaps this can be done for his PR in to your devel?

lohedges (changes requested):
This is a first pass review. I'll try to go through things again this afternoon / tomorrow and will update with anything I've missed.

It's a bit tricky to review since this also includes parital edits from @annamherz's Frankenstein [feature-analysis](https://github.com/michellab/BioSimSpace/tree/feature-analysis) branch and some WIP from @fjclark (has this evolved now, e.g. in the SOMD ABFE PR to EXS `devel`?).

xiki-tempula:
@lohedges Without looking through the comments.

So I have not included anything from @annamherz's Frankenstein [feature-analysis](https://github.com/michellab/BioSimSpace/tree/feature-analysis) branch. I have made my own attempt at changing the analysis function as I'm more familiar with alchemlyb.

There is some WIP from @fjclark I could remove them if it makes your review easier. It is just that I fear that it will break https://github.com/Exscientia/BioSimSpace/pull/19

lohedges:
Ah, apologies, I was under the impression that you had merged across some of Anna's alchemlyb edits early on, I hadn't realised that you'd done your own edits. This might make things tricky when Anna's work is merged in, but we can deal with that then.

I'd leave Finlay's bit if it's working in this version. You can ignore comments on this and I'll try to copy them over to the other PR. It might be confusing as there will no longer be a diff to refer to, though.

Cheers.

lohedges:
Alternatively, I could try reviewing @fjclark's PR first, so that this PR would be consistent once that's merged.

xiki-tempula:
@lohedges For me, it might be better if this PR gets merged first as we are preparing another several big PR now.

lohedges:
No problem, whatever is preferable. Ultimately this is your sandpit, so what goes in there is up to you. Any comments / suggestions made in review here are mostly for the benefit of when it comes to porting those updates into `devel`.

lohedges:
> So I have not included anything from @annamherz's Frankenstein [feature-analysis](https://github.com/michellab/BioSimSpace/tree/feature-analysis) branch.

It looks like this came in during the early sandpit synchronisation, i.e. when some of the edits from the [feature-amber-fep](https://github.com/michellab/BioSimSpace/tree/feature-amber-fep) branch were merged back into yours.

No worries about this, just a note to myself (everyone) that there might be some conflicts whenever [feature-analysis](https://github.com/michellab/BioSimSpace/tree/feature-analysis/python/BioSimSpace) is merged in, since it now comes from a different base.

.github/workflows/Sandpit_exs.yml
xiki-tempula: Sorry what do you mean? The `alchemlyb` is installed by conda?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
lohedges: I've had a quick look and it appears that _all_ of the dependencies for `alchemlyb` are available on all supported architectures, so you should just be able to import it directly, i.e. no need to use `_try_import`.

lohedges: Run requirements for `alchemlyb` are:

```
  run:
    - matplotlib-base
    - numpy
    - pandas >=1.4
    - pymbar >=3.0.5,<4
    - python >=3.8
    - scikit-learn
    - scipy
```

xiki-tempula: So I would assume that alchemlyb is a dependency and there is no need to wrap it in a `try/except` block?

lohedges: Yes, that's right. It's not a dependency (yet) in the main part of the repository, i.e. not in our `requirements.txt`, so (for now) you'll still need to remember to install in manually, as you are doing in your CI script.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
xiki-tempula: I thinking passing 9 parameters explicitly is a bit extreme.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
xiki-tempula: Why would there be space?

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
lohedges: I wasn't implying all 9 independently, rather the three keys in the dictionary, i.e. `anchor_points`, `equilibrium_values`, and `force_constants`.

From my comment later, it looks like this object isn't created directly by the user anyway (it's created for them by the restraint generation code) so it probably doesn't matter too much.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
lohedges: There might not be, but you'll be surprised what people type. No worries if not, just this is done in a lot of other places in the code in order to sanitise a string and catch easy typos. For example I have sometimes come across white-space issues when the parameter is passed in from a configuration file, or as a command-line parameter to a `Node`.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
xiki-tempula: Sorry, what is the problem here?

lohedges: Apologies, I see what you are doing here. Just looking at the commit diff it is unclear that these parts of the documentation aren't expanded from variables elsewhere. Looking at the full file makes things clear. Since it's not possible for the user to set the `ligand_resname` (it's taken directly from the decoupled molecule) perhaps it's useful to highlight what it means and that it's not an adjustable user parameter. (I guess they can change it if they call the private analyse methods directly, but it's not something that they need to do.)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
xiki-tempula: So the problem is how to pass the restraint information.
My current plan is that the user would generate the `Restraint` using `RestraintSearch` and then pass the information to process. I don't think user is able to generate the restraint by themselves as unless the restraint are carefully generated, it will almost always explode.

Yes, I can make Restraint into a private class.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
xiki-tempula: Not addressed as it is Finlay's code.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
lohedges: Great, that was exactly what I was thinking :+1:

fjclark: Would it be good to allow the user to supply the restraint information in order to reproduce previous simulations/ results from the literature? I've assumed that's how it would work, and I've added  some minimal checks in my branch (https://github.com/Exscientia/BioSimSpace/commit/fdf4f46dd7e86cf6c2d9cf55f865241d4c70111d , https://github.com/Exscientia/BioSimSpace/commit/3a3f3703cb639c54574bfb8ead0758bd6d20054f) to ensure that the user-supplied restraints are sensible.

python/BioSimSpace/Sandpit/Exscientia/Protocol/_free_energy_mixin.py
xiki-tempula: > We currently implicitly depend on pandas via other dependencies, so would need to add a hard dependency if used.

I think in these days, it is very hard to escape pandas in the dependency.

Fo other packages, I think it is fair that we should make the input as general as possible and not rely on specific package, but I think we do expect user to familiar with some basic packages such as numpy, pandas or matplotlib. I'm not quite sure how many users could benefit from using a dictionary instead of a pandas dataframe. 

On the other hand, If we tell the user that we are expecting a pd.dataframe, the user will know what kind of data structure that we are expecting without needing to read the documentation. If we tell user that we are expecting a dictionary, the user will need to read the description to understand how to format the input. 

There is also the benefit that the user could look at the input and have a rough idea of what is happening behind curtain and the benefit of keeping the input and internal data structure same.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_relative.py
xiki-tempula: Sorry, what do you mean here?

lohedges:
Sorry, I see this part has been removed. Just a general comment about
variable names :-)

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
xiki-tempula: Sorry, I'm struggling a bit on how to do this, currently, I did this with
```
_MDRestraintsGenerator = _try_import("MDRestraintsGenerator",
                                     install_command='pip install MDRestraintsGenerator')

from _MDRestraintsGenerator import search as _search
from _MDRestraintsGenerator.restraints import FindBoreschRestraint as _FindBoreschRestraint
```
This will fail with
```
  File "/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3433, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-90e5ed732635>", line 1, in <module>
    from _MDRestraintsGenerator import search as _search
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_bundle/pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
ModuleNotFoundError: No module named '_MDRestraintsGenerator'
```
The _MDRestraintsGenerator seems to be the right module though
```
_MDRestraintsGenerator
Out[4]: <module 'MDRestraintsGenerator' from '/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.8/site-packages/MDRestraintsGenerator/__init__.py'>
```
`from MDRestraintsGenerator import search as _search` would work fine.

lohedges: You can't import and rename, then use the renamed module to import from. It's just a local variable name, so isn't used by the import system.

I think something like the following is what we want:

```
_MDRestraintsGenerator = _try_import("MDRestraintsGenerator",
                                     install_command='pip install MDRestraintsGenerator')
                                     

if _have_imported(_MDRestraintsGenerator):
    from MDRestraintsGenerator import search as _search
    from MDRestraintsGenerator.restraints import FindBoreschRestraint as _FindBoreschRestraint
```

This way, we only do the `from X import Y` bit if the `_try_import("X")` bit worked.

python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
xiki-tempula: Sorry I'm not very familiar with this API.
So for a system that has decoupled molecule, if I do `self.search("property decouple")`, I got 
```
  File "/Users/zwu/miniconda3_x86/envs/exs-orionbssfloes/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3433, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-10-b8fc0224423d>", line 1, in <module>
    self.search("property decouple")
  File "/Users/zwu/bitbucket/BioSimSpace/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py", line 983, in search
    raise ValueError(msg) from None
ValueError: 'Invalid search query: 'property decouple' : 'Nothing matched the search.'
```

The decoupled molecule is there as if I do `self._sire_object[0].has_property('decouple')`, I got True.

```
>>> self._sire_object[0].property('decouple')
Properties(
    LJ => [ 1,0 ],
    intramol => True,
    charge => [ 1,0 ]
)
```

lohedges: Hmm, that's strange. For example, if I do the following:

```python
import BioSimSpace as BSS

s = BSS.IO.readMolecules("amber/ala/*")

# Extract the first molecule and give it some dummy property.
m = s[0]
m._sire_object = m._sire_object.edit().setProperty("test", 1.0).molecule().commit()

# Update the molecule in the sytem.
s.update(m)

# Now search for molecules with the matching property.
s.search("property test").molecules()[0]
<BioSimSpace.Molecule: nAtoms=22, nResidues=3>
```

xiki-tempula: So in this case, `decouple` is not a `int` but a dictionary with a couple keys.
```
Properties(
    LJ => [ 1,0 ],
    intramol => True,
    charge => [ 1,0 ]
)
```
Would this work as well?

lohedges: I've just tried and it does indeed fail when using a dictionary. I'll raise this as a Sire issue since there's no reason why it shouldn't work the same. Leave things as they are for now and we can update when it's working.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
lohedges: Okay, will note to mention this in the other PR.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint_search.py
lohedges: That's a good point. I guess a dictionary for the parameters makes sense here since it's a single object that could be pickled easily.

python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
chryswoods: Lester and I have talked about this - the current property search only supports simple properties (convertible to number or string). This is an excellent use case for supporting more complex properties. I am looking at adding this now to the `feat_2023_0` branch in Sire (as well as me making the cursor work with setting list and dictionary properties). Lester will update this once this is finished and will update once `feat_2023_0` is merged in and it is working.

chryswoods: That was quicker and easier than I thought. I've added the search for molecules that have a matching property in [this commit](https://github.com/michellab/Sire/commit/941603ba2dd5ef8fa39e8859e364c6fc8858bd8f) on `feat_2023_0`. It is quite self-contained (just some edits to `idengine.cpp` plus adding in a new file for search unit tests). You could merge this single commit into `devel` if you want access to this earlier.

I also fixed the error when assigning list/dict properties in the [prior commit](https://github.com/michellab/Sire/commit/244bca454f8d76ac0f946e8f6df4ee2e67eb2e6f). This now works as you would expect, e.g. `cursor["list_prop"] = [1, 2, "cat"]`.

xiki-tempula:
@lohedges Many thanks for the comments, I have addressed all the comments in the latest commit.

lohedges (approved):
Thanks for the edits. Looks good to me. I'll submit my own PR once the functionality to directly search for the `decouple` property has been merged into Sire.

Cheers.

PR #365, opened 2022-07-22
Feature analysis

annamherz:
Hello! This is the PR relating to the updated analysis in BSS. I have tested this for all three engine outputs via the following, with work_dir being the directory of the run and leg being either the bound or free leg:

`pmf, overlap_matrix = BSS.FreeEnergy.Relative.analyse(f"{work_dir}/{leg}", 'MBAR')`

and this also works if the estimator is set to TI.
This proceeds by first extracting the data from the output files depending on the engine, and then carrying out the equilibrium detection, statistical inefficiency, and auto MBAR or TI using alchemlyb.

I also tested it for starting FreeEnergy runs with SOMD and directly analysing the outputs in BSS:
```

bound_run = BSS.FreeEnergy.Relative(
    system_bound,
    freenrg_protocol,
    engine=f"{engine_query}",
    work_dir=f'{work_dir}/bound'
)
bound_run.run()

free_run = BSS.FreeEnergy.Relative(
    system_free,
    freenrg_protocol,
    engine=f"{engine_query}",
    work_dir=f'{work_dir}/free'
)
free_run.run()

pmf_bound, overlap_bound = bound_run.analyse()
pmf_free,  overlap_free  = free_run.analyse()
free_nrg_binding = BSS.FreeEnergy.Relative.difference(pmf_bound, pmf_free)

```
if TI is to be used with .analyse(), it needs to be set during the BSS.FreeEnergy.Relative stage, as otherwise MBAR is the default. Otherwise it can also be used using the whole `BSS.FreeEnergy.Relative.analyse(f"{work_dir}/{leg}", 'TI') ` for that run and that works fine.

Checking the overlap for MBAR works using:

`free_run.check_overlap()`

or
```
BSS.FreeEnergy.Relative.check_overlap(free_overlap, estimator='MBAR')
BSS.FreeEnergy.Relative.check_overlap(free_overlap) # or this as MBAR is the default
```

I also added the plotting from alchemlyb for the overlap and dhdl as well, eg:

`
free_run.plot()`

or

`overlap_plot = BSS.FreeEnergy.Relative.plot(overlap_free, estimator='MBAR', work_dir=f'{work_dir}/free')`

When no work dir is defined, the output isn't saved so for jupyter notebooks it would only show it. The estimator must match the estimator of the anaylse used, as otherwise the data format is not correct for MBAR/TI. If no estimator is set, it will infer the type of data and if MBAR/TI plotting should be used.

Please let me know any things I should change, or if there are any code mistakes, or any ideas for improvements!


lohedges:
Thanks, @annamherz! I should be able to take a look at this in detail early next week.

lohedges:
Sorry for the delay with this. I'll try to make some comments today/tomorrow. Don't worry about making any edits before the workshop, we can deal with them afterwards when we think about merging things into devel. It might be the case the workshop has highlighted some other things that need editing/improving too.

Cheers.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -795,23 +1523,23 @@ def _initialise_runner(self, system):
...
+                    process._protocol = self._protocol
+                    process._work_dir = new_dir
+                    process._std_out_file = new_dir + "/somd.out"
+                    process._std_err_file = new_dir + "/somd.err"
+                    process._rst_file = new_dir + "/somd.rst7"
+                    process._top_file = new_dir + "/somd.prm7"
+                    process._traj_file = new_dir + "/traj000000001.dcd"
+                    process._restart_file = new_dir + "/latest.rst"
+                    process._config_file = new_dir + "/somd.cfg"
+                    process._pert_file = new_dir + "/somd.pert"
```
lohedges, line 1537: I'm not sure if this was the result of a whitespace trim? The previous alignment was intended to aid readability

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -417,7 +675,292 @@ def _analyse(self):
...
+
+           Returns
+           -------
+
+           pmf : [(float, :class:`Energy <BioSimSpace.Types.Energy>`, :class:`Energy <BioSimSpace.Types.Energy>`)]
+               The potential of mean force (PMF). The data is a list of tuples,
+               where each tuple contains the lambda value, the PMF, and the
+               standard error.
+
+           overlap : [ [ float, float, ... ] ]
```
lohedges, line 762: Just to check... with the new analysis, is the overlap always returned as a numpy matrix? If so, can the docstrings be updated appropriately, i.e. to `overlap : numpy.matrix`.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
         # Now call the staticmethod passing in both PMFs.
         return Relative.difference(pmf, pmf_ref)
 
+    @staticmethod
+    def check_overlap(overlap, estimator="MBAR"):
```
lohedges, line 1282: For agreement with the current API naming conventions, this should be `checkOverlap`.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+           overlap : [ [ float, float, ... ] ], numpy.matrix
+               The overlap matrix. This gives the overlap between lambda windows.
+
+           estimator : str
+               Must be "MBAR" for checking the overlap matrix.
+
+           Returns
+           -------
+
+           Message - gives a warning if the overlap is less than 0.03 for any off-diagonal.
```
lohedges, line 1297: Can this return nothing and throw an exception, or return a boolean, e.g. with False being poor overlap. I imagine a workflow where the user with have call `checkOverlap` in a script and decide whether to proceed (or submit more jobs) based on the result.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -417,7 +675,292 @@ def _analyse(self):
...
+
+           Returns
+           -------
+
+           pmf : [(float, :class:`Energy <BioSimSpace.Types.Energy>`, :class:`Energy <BioSimSpace.Types.Energy>`)]
+               The potential of mean force (PMF). The data is a list of tuples,
+               where each tuple contains the lambda value, the PMF, and the
+               standard error.
+
+           overlap : [ [ float, float, ... ] ]
```
lohedges, line 841: I guess `overlap` doesn't make sense here, so we have some inconsistency in what is returned depending on what estimator is used. Is the TI gradient object the same, i.e. a matrix of floats, or different?

It might be worth thinking about having separate `analyseMBAR` and `analyseTI` methods, rather than getting different output based on the estimator. As things stand, we would need to update _all_ docstrings to make it clear that what is returned in each case.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+                ax.figure.savefig(
+                    f"{work_dir}/overlap_MBAR.png", bbox_inches='tight', pad_inches=0.0)      
+
+        elif estimator == 'TI':
+            if not isinstance(overlap, _alchemlyb.estimators.ti_.TI):
+                if not isinstance(overlap, list):
+                        raise TypeError("'overlap' must be of type 'alchemlyb.estimators.ti_.TI' for 'TI'.\
+                                          This is obtained from running analysis using estimator='TI'.\
+                                          Alternatively, it may also be of type 'list'.")
+
```
lohedges, line 1403: Referring back to my earlier comment, I can now see what type `overlap` will be in the case of a TI estimator. I think it would be best to discuss ways to get around this issue, since the use of `overlap` is likely to lead to confusion here. I see that the output can be of type `_alchemlyb.estimators.ti_.TI`, or `list`. For the latter, you'd probably also need to validate the contents of the list, since Python lists can contain anything.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+        """
+
+        # Calculate the overlap for this object.
+        _, overlap = self.analyse()
+
+        # Now call the staticmethod passing in the overlap and the work_dir of the run.
+        return Relative.check_overlap(overlap, estimator=self._estimator)
+
+    @staticmethod
+    def plot(overlap, estimator=None, work_dir=None):
```
lohedges, line 1348: Just to note that we already have functionality to plot the overlap matrix, which was adapted from @JenkeScheen's code and can be found [here](https://github.com/michellab/BioSimSpace/blob/082f4433a0f77fcecb3696d4b8a3708c9bd32332/python/BioSimSpace/Notebook/_plot.py#L464). I'd rather not have two bits of functionality to do the same thing, so will probably remove the old one.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+
+            # use the alchemlyb functionality to plot the dhdl
+            ax = _plot_ti_dhdl(overlap)
+            ax.set_title(f"dhdl plot")
+            
+            if work_dir is not None:
+                ax.figure.savefig(
+                    f"{work_dir}/dHdl_TI.png")
+
+        return ()
```
lohedges, line 1412: The method implies that you are returning the plot object, which would allow the user to edit it. I think it's fine to return nothing (just remove this and the docstring).

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+                        raise TypeError("'overlap' must be of type 'numpy.matrix' for 'MBAR'.\
+                            This is obtained from running analysis using estimator='MBAR'.")
+
+            # use the alchemlyb functionality to plot the overlap matrix
+            ax = _plot_mbar_overlap_matrix(overlap)
+            ax.set_title(f"overlap matrix")
+
+            if work_dir is not None:
+                ax.figure.savefig(
+                    f"{work_dir}/overlap_MBAR.png", bbox_inches='tight', pad_inches=0.0)
```
lohedges, line 1395: There is no mention that the plot will be saved if a `work_dir` is specified. The `work_dir` also isn't validated, so this will error if it doesn't exist or the path is nonsensical.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -696,6 +1278,159 @@ def _difference(self, pmf_ref):
...
+                raise TypeError("Data type for estimator = 'None' could not be inferred / does not match allowed data types.")
+
+        if estimator == "MBAR":
+            if not isinstance(overlap, _np.matrix):
+                        raise TypeError("'overlap' must be of type 'numpy.matrix' for 'MBAR'.\
+                            This is obtained from running analysis using estimator='MBAR'.")
+
+            # use the alchemlyb functionality to plot the overlap matrix
+            ax = _plot_mbar_overlap_matrix(overlap)
+            ax.set_title(f"overlap matrix")
```
lohedges, line 1391: What would happen when this method is called when the user isn't in an interactive environment? As in [BioSimSpace.Notebook._plot.py](https://github.com/michellab/BioSimSpace/blob/devel/python/BioSimSpace/Notebook/_plot.py#L464), you'll probably want to check for an interactive session before plotting.

lohedges:
Just added some comments. This is still a WIP so I'll add more if I think of anything.

Thanks again for this. It will be good to catch up after the workshop to discuss any further refinements based on feedback.

recipes/biosimspace/template.yaml
```diff
@@ -21,6 +21,7 @@ requirements:
     - configargparse
     - fkcombu
     - lomap2
+    - alchemlyb
```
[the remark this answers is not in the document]
annamherz: okay sorted!

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -29,26 +29,40 @@
 __all__ = ["Relative", "getData"]
 
 from collections import OrderedDict as _OrderedDict
+from email.policy import default
```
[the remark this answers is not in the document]
annamherz: I'm not really sure why that was there actually, have gotten rid of it for now.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -191,6 +216,21 @@ def __init__(self, system, protocol=None, work_dir=None, engine=None,
                     raise NotImplementedError("GROMACS currently only supports the 'full' perturbation "
                                               "type. Please use engine='SOMD' when running multistep "
                                               "perturbation types.")
+                self._exe = _gmx_exe
+            
+            elif engine == "AMBER":
+                _warnings.warn(
+                    "In this branch, AMBER is currently only available for analysis of AMBER output.")
```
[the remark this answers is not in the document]
annamherz: I think I initially put this in so when the analysis is merged the amber functionality is present for later on, but as the actual running of Amber simulations for relative isn't yet in the devel it would throw a warning if setting up and running those is tried.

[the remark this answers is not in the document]
lohedges: Ah, I see, that makes sense. Sorry for misinterpreting it. Yes, please leave this there for now.

python/BioSimSpace/FreeEnergy/_relative.py
annamherz: Ah yes that must have been the automatic formatting, I've changed it back.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -417,7 +675,292 @@ def _analyse(self):
...
+        Args:
+            data : dataframe
+            Dataframe of extracted dHdl or u_nk data.
+
+        Returns
+        -------
+
+            processed_data : dataframe
+            Dataframe of dHdl or u_nk data processed using automated equilibration
+            detection followed by statistical inefficiency.
```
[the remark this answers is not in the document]
annamherz: I've changed it to private for now, but after some discussion with @jmichel80 I think it might be good for the user to have more input about how their data is preprocessed. I'm not sure what the best way would be to go about this, as the preprocessing is quite hidden in the analyse part right now (as a public API? or via extra options?), but I think this would also be good to discuss more after the workshop.

python/BioSimSpace/FreeEnergy/_relative.py
```diff
@@ -353,14 +405,217 @@ def getData(self, name="data", file_link=False, work_dir=None):
...
+        outfile : str
+            Path to simfile.dat file to extract data from.
+        T : float
+            Temperature in Kelvin at which the simulations were performed;
+            needed to generated the reduced potential (in units of kT)
+
+        Returns
+        -------
+        data : DataFrame
+            Reduced potential for each alchemical state (k) for each frame (n).
```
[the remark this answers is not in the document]
annamherz: I've just changed this but after running some test analysis I've realised a problem -

```
        function_glob_dict = {
            "SOMD": (Relative._somd_extract_u_nk),
            "GROMACS": (_gmx_extract_u_nk),
            "AMBER": (_amber_extract_u_nk)
        }

        # Extract the data.
        func = function_glob_dict[engine]
        try:
            u_nk = [func(x, T=t) for x, t in zip(files, temperatures)]
        except:
            raise _AnalysisError(
                "Could not extract the data from the provided files!")
```

In this part of the code, as T is what is used for the amber and gromacs which are part of alchemlyb, changing it to temperature makes it inconsistent with these.

python/BioSimSpace/FreeEnergy/_relative.py
annamherz: I've tried to make it a bit more clear for now by changing it to dHdl (and also put the correct type which is `alchemlyb.estimators.ti_.TI`), but I agree and I think it might be cleaner to have different analyse for each. Also, currently the estimator is set during the init so that if an object has been run it is still able to be self analysed as before. However, this also currently limits this self analyse to only one estimator (unless that property is modified before analysing with the other estimator, but this doesn't feel very intuitive).  I haven't really been using it like this as I usually have run any perturbations outside of BSS and therefore can just choose which estimator to use when doing the full `BSS.FreeEnergy.Relative.analyse()` , but yea I think having seperate analysers for each might be a good idea as it would also make this part more clear.

lohedges: Yes, there's probably no need to specify the estimator when the `FreeEnergy.Relative` object is initialised, rather have it solely as an option to the `analyse` method (with a default). I'll have a think about separate analysers, since I agree that it is the clearer solution. Perhaps we could have, e.g. `analyseMBAR` and `analyseTI`, with the regular `analyse` method simply calling those, and left in place for backwards compatibility.

python/BioSimSpace/FreeEnergy/_relative.py
annamherz: Oh yea! I was trying to keep in consistent so it uses the alchemlyb library and forgot about that.

python/BioSimSpace/FreeEnergy/_relative.py
annamherz: When I called it in a non interactive environment it only saves the plot as an image if the work directory is specified, but otherwise does nothing. In that case would I still need to check for an interactive environment?

annamherz:
Thank you so much for all the feedback 😊 I'm also currently doing some large scale analysis to check the reproducibility between the statistical inefficiency / autoequilibration detection and other analysis protocols, so I think after this I'll have a more clear idea of what the best way to handle these with BSS would be as well.

python/BioSimSpace/FreeEnergy/_relative.py
lohedges: That's okay, if it can still write the image non-interactively, then it makes sense. I've just come across certain cases where certain plotting functionality can crash/hang if run non-interactively, or without an X session.

annamherz:
Another potential further thing for analysis, that is implemented for now very roughly in the workshop for the sake of the absolute analysis (6aaae504d04d1d3bdef0b3bdf69caccf6ace2017 , fixed in d084e58658878811f48c77cda22f5dcba00da6e8, warnings in 6ac77fb5496a9e3a7b19d839e593b3ad3a1832bd) is a way to use the the previous ('native') analysis for SOMD so that previous results using older BSS can still be reproduced. I'll still test this some more and I think for the main analysis feature if we keep it the implementation of it would need some more thought as I don't think the way I've put it in right now is ideal.

annamherz:
> Another potential further thing for analysis, that is implemented for now very roughly in the workshop for the sake of the absolute analysis ([6aaae50](https://github.com/michellab/BioSimSpace/commit/6aaae504d04d1d3bdef0b3bdf69caccf6ace2017) , fixed in [d084e58](https://github.com/michellab/BioSimSpace/commit/d084e58658878811f48c77cda22f5dcba00da6e8), warnings in [6ac77fb](https://github.com/michellab/BioSimSpace/commit/6ac77fb5496a9e3a7b19d839e593b3ad3a1832bd)) is a way to use the the previous ('native') analysis for SOMD so that previous results using older BSS can still be reproduced. I'll still test this some more and I think for the main analysis feature if we keep it the implementation of it would need some more thought as I don't think the way I've put it in right now is ideal.

I've merged the changes made for the workshop so that they're together with the analysis stuff.

> I'm also currently doing some large scale analysis to check the reproducibility between the statistical inefficiency / autoequilibration detection and other analysis protocols, so I think after this I'll have a more clear idea of what the best way to handle these with BSS would be as well.

After doing this, it seems like the statistical inefficiency / subsampling (for the SOMD native) does not improve agreement of the results with the experimental, and also makes the alchemlyb and native analysis methods less consistent with each other, so I've turned it off for now as well.

PR #374, opened 2022-09-01
Set ipywidgets<8 to Avoid Issues with NGLView

fjclark:
Hello,

Recent changes to ipywidgets have [broken NGLView](https://github.com/nglviewer/nglview/issues/1032) - ensuring version <8 avoids any issues.

Thanks,
Finlay

### DIFF
--- a/docker/conda/recipe/template.yaml
+++ b/docker/conda/recipe/template.yaml
@@ -20,6 +20,7 @@ requirements:
     - mdtraj
     - mdanalysis
     - nglview
+    - ipywidgets<8
     - openff-toolkit-base
     - parmed
     - pydot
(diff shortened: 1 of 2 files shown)


lohedges:
Sorry, missed your link to the bug report. I think we can use this workaround for now and update if/when NGLView fix the issue upstream.

Cheers

fjclark:
Great. There don't seem to be any issues with interactive nodes with GUI elements - just with NGLView.

Cheers

lohedges:
Thanks for confirming. We only use pretty basic widgets so I'm glad it's unaffected. Should hopefully be able to merge this in tomorrow.

PR #382, opened 2022-09-14
Use the Free Energy Analysis pipeline from alchemlyb

xiki-tempula:
Use the Free Energy Analysis pipeline from alchemlyb


jmichel80:
Is this PR not going ahead ? it would good to check with @annamherz and @xiki-tempula  have converged on the same best practices for free energy estimation with alchemlyb

PR #394, opened 2022-10-21
Fix offset when injecting position restraints into GROMACS top file

lohedges:
This PR fixes the offset used when injecting the inclusion of external `posre*.itp` files into a GROMACS topology. The logic used previously meant that the offset was incorrect, meaning that the `#include ...` directive was injected at the wrong position when more than one molecule type was present in the topology, potentially overwriting existing records for bonded terms.

@msuruzhon: I appreciate that this fix is applied to an outdated version of the sandpit, but at least you can copy into your development version to test. If it's easier, I could also apply to the sandpit on the `merge-exs-devel` branch, where I am attempting to synchronise things for the 2023 Sire API updates.

Cheers.


python/BioSimSpace/Process/_gromacs.py
```diff
@@ -2255,7 +2262,7 @@ def _add_position_restraints(self, config):
 
                 # Write the updated topology to file.
                 with open(self._top_file, "w") as file:
-                    for line in top_lines:
+                    for line in new_top_lines:
                         file.write("%s\n" % line)
 
             # A user-defined list of atoms indices.
```
msuruzhon: Is this issue also not present in this branch of the conditional?

python/BioSimSpace/Process/_gromacs.py
```diff
@@ -2194,6 +2194,12 @@ def _add_position_restraints(self, config):
                 # The number of restraint files.
                 num_restraint = 1
 
+                # Store the number of molecule types in the top file.
```
msuruzhon: Would be nice to add a test for this I think because there are quite a few operations going on in this function.

msuruzhon:
Thanks @lohedges I have a couple of comments and I am also going to test it and I will approve once I have tested it on the original system I tried it on (more complex than what I shared for reproduciility). Many thanks.

python/BioSimSpace/Process/_gromacs.py
lohedges: Ah yes, good spot. Will update this now.

python/BioSimSpace/Process/_gromacs.py
lohedges: Yes, I agree that it would be good to add a test. I'll try to come up with a small example. (Could use yours, but the files are quite large.)

python/BioSimSpace/Process/_gromacs.py
lohedges: This should now be resolved. When I get a chance I might try to consolidate the two branches with a function call. The logic used to be quite a bit more different, but I managed to adjust things a while ago so the loops are effectively the same, although use different data structures.

msuruzhon:
@lohedges I have tested it on a few systems and it seems to work for me. Feel free to merge once you add some tests. Many thanks.

lohedges:
Thanks for confirming. I'm just trying to devise a test to cover all possible locations for the position restraint file, e.g. when the restraint applies to one, more than one, or all molecules, and when the order of molecules in the file is changed, e.g. in which case the restraint might only apply to the last molecule.

PR #397, opened 2022-11-11
Make sure ordering is preserved when a molecule is replaced

lohedges:
This PR ensures that molecule ordering is preserved when a molecule at a given index is _replaced_ by another, i.e. by a molecule with a different `MolNum`. Previously we removed the old molecule, then added the new one, which caused a re-ordering of the underlying container, i.e. the new one would be at the end.

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
@@ -108,11 +108,11 @@ class Molecule(_SireWrapper):
 
     def __str__(self):
         """Return a human readable string representation of the object."""
-        return "<BioSimSpace.Molecule: nAtoms=%d, nResidues=%d>" % (self.nAtoms(), self.nResidues())
+        return "<BioSimSpace.Molecule: number=%d, nAtoms=%d, nResidues=%d>" % (self.number(), self.nAtoms(), self.nResidues())
 
     def __repr__(self):
         """Return a string showing how to instantiate the object."""
-        return "<BioSimSpace.Molecule: nAtoms=%d, nResidues=%d>" % (self.nAtoms(), self.nResidues())
+        return "<BioSimSpace.Molecule: number=%d, nAtoms=%d, nResidues=%d>" % (self.number(), self.nAtoms(), self.nResidues())
 
     def __add__(self, other):
         """Addition operator."""
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_system.py
@@ -544,22 +544,16 @@ class System(_SireWrapper):
 
         # The molecule numbers don't match.
         else:
-            # Store the original molecule number.
-            mol_num = self[index]._sire_object.number()
+            # Create a copy of the system.
+            system = self.copy()._sire_object
 
-            # Store a copy of the passed molecule.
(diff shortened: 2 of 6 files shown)


msuruzhon (approved):
Thanks @lohedges I tested this on my larger test suite that was failing and it seems to pass now as well so this seems to be good to merge.

msuruzhon:
Ah, I spoke too soon, I get a different test failing now because of reordering. Will share once I manage to isolate the problem.

lohedges:
No problem. Let me know when you've found an example and I'll debug.

Cheers.

msuruzhon:
Okay so I think this might be related to the search function in Sire?

Using the previously shared pickled system:

```python
pertmol_idxs = [i for i, molecule in enumerate(system.getMolecules()) if molecule.isPerturbable()]
pert_mols = system.getPerturbableMolecules()
assert system[pertmol_idxs[0]].nAtoms() == pert_mols[0].nAtoms()
```

msuruzhon:
So when I try with the attached system and the following script:

```python
system = pickle.load(open(f"Broken_system.pickle", "rb"))
pertmol_idxs = [i for i, molecule in enumerate(system.getMolecules()) if molecule.isPerturbable()]
pert_mols = system.getPerturbableMolecules()
assert system[pertmol_idxs[0]].nAtoms() == pert_mols[0].nAtoms()
```

I get a discrepancy, which is due to `getPerturbableMolecules` returning the incorrect ordering, presumably because of the underlying call to the Sire search function?
[Archive.zip](https://github.com/michellab/BioSimSpace/files/10002468/Archive.zip)

lohedges:
Okay, it looks like the issue here is that the search is being returned in `MolNum` order, so it isn't corresponding to the system, where the `MolNum`s are now potentially out-of-order, e.g. if you've replaced a molecule. In order to fix this, I think I am going to have to use my previous approach, i.e. renumber the replacement molecule when it is inserted into the system at a specific index, i.e. match the existing `MolNum`. However, this might open a whole can of worms, since you'll no longer be able to do things like checking a molecule is _in_ a system, since the number won't match.

```python
# Search is returned in MolNum order.
system._sire_object.search("perturbable")
Out[7]:
SelectResult( size=2,
0 : Molecule( Merged_CL:10597 num_atoms=1 num_residues=1 )
1 : Molecule( Merged_Molecule:10598 num_atoms=45 num_residues=1 )
)

# However, the system contains molecules that are out-of-order.
pert_nums = [mol.number() for mol in system if mol.isPerturbable()]
[10598, 10597]
```

lohedges:
Could you try the latest edits? I'll discuss this with @chryswoods, since it might be preferable to return searches in `MolIdx` order, rather than enforcing that `MolNums` are ascending within a system.

lohedges:
Note my test is now invalid given the most recent edits. Still figuring out if it's best to fix the search.

msuruzhon:
Ah, let me know when this is ready for me to test then thanks.

lohedges:
Okay, this should now be good to test. I've reverted the molecule replacement behaviour so that the replacement molecule keeps its `MolNum`, rather than taking the one of the molecule that it replaces. Instead, @chryswoods fixed the molecule ordering issue when searching. Molecules are now always returned in `MolIdx` order. Before it was essentially random since the underlying container made no assumptions about ordering.

msuruzhon:
This works for me now @lohedges, all tests pass. Many thanks!

lohedges:
Ah brilliant, thanks for letting me know. I'll merge this in once I've fixed the issues related to the ABFE merge. Thanks again for catching this.

PR #398, opened 2022-11-14
Fix CI issues following EXS sandpit merge

lohedges:
This PR fixes some CI issues related to the recent EXS sandpit merge [here](https://github.com/michellab/BioSimSpace/pull/363):

* Recent versions of `alchemlyb` aren't available for Python 3.7. As such, we now use `_try_import` to make sure BioSimSpace can still be imported correctly.
* Tests that require `ambertools` or `gromacs` are now marked to be skipped if these aren't installed.
* The CI runner script that was introduced with the 2023 edits incorrectly pulls from michellab for _all_ pull requests, i.e. it only uses the `head_ref` for that repo. This is why the above issues weren't caught by the CI in the previous PR. I've switched to using `github.repositoryUrl`, although I'll need to test that this works (it uses the default Git protocol, not HTTPS, which hangs) in this PR. I'll try different strategies if it doesn't work.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # BioSimSpace runtime requirements.
 
-sire ~=2023.0.0   # this matches >= 2023.0.0 && < 2023.1.0
+sire ~=2023.0.1   # this matches >= 2023.0.1 && < 2023.1.0
 
 alchemlyb ; python_version >= '3.8'
 configargparse
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 
 sire ~=2023.0.0   # this matches >= 2023.0.0 && < 2023.1.0
 
-alchemlyb
+alchemlyb ; python_version >= '3.8'
 configargparse
 kcombu_bss
 ipywidgets<8
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -55,6 +55,9 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
 #
+      - name: Use HTTPS git
+        run: git config --global url.https://github.com/.insteadOf git@github.com:
+#
       - name: Clone the devel branch (push to devel)
         run: git clone ${{ github.repositoryUrl }}
         if: github.event_name != 'pull_request'
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -56,11 +56,11 @@ jobs:
           use-mamba: true
 #
       - name: Clone the devel branch (push to devel)
-        run: git clone https://github.com/michellab/BioSimSpace
+        run: git clone ${{ github.repositoryUrl }}
         if: github.event_name != 'pull_request'
 #
       - name: Clone the feature branch (pull request to devel)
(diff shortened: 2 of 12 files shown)


xiki-tempula:
Do we still need to support py3.7? https://numpy.org/neps/nep-0029-deprecation_policy.html

lohedges:
Hopefully we can deprecate it soon. We were actually only using it since you were, but it looks like you've already updated. Let me know what version requirements you have. I'll then check with other users and come up with a deprecation strategy.

xiki-tempula:
Currently we are using only using py3.8.

lohedges:
It looks like `github.repositoryUrl` fails, i.e. it's waiting for password input. I'll see if there's a way to generate the HTTPS URL instead.

xiki-tempula:
I'm happy with the changes, do we expect the build to succeed?

lohedges:
Yes, it should work. I just need to figure out how to set the URL correctly.

lohedges:
Okay, this is now passing. A few more edits were required in relation to tests that weren't skipped when optional dependencies were missing. I've locally updated to use the new Sire property searching functionality and will push a few minor edits once the Sire CI build has completed.

xiki-tempula (approved):
LGTM

lohedges:
Failures are because the CI is pulling in an outdated version of Sire during the build. Presumably it is doing so because of the additional `ambertools` and `gromacs` host requirements used in the build, which likely pull in different package versions.

I'll need to create a 2023.0.1 release of Sire to fix this. Bah.

lohedges:
I realised that the 2023 update meant that BioSimSpace was building against packages using the `main` label on our Anaconda channel, so was always using an outdated version of Sire. I've fixed this, but the incorrect package (build 21, not 31, which is the latest) is pulled in for Python 3.7. This is the build prior to the switch to using `ambertools` and `gromacs` as host requirements for Sire.

I'll create a new Sire release so that we can build against that instead. (Alternatively, we could drop Python 3.7 support. It seems that some of our dependencies are no longer being built for this variant, so environments are becoming hard to resolve.

xiki-tempula:
Might be easier to drop 3.7

lohedges:
It turns out that py310 is problematic on macOS and Windows, so I've created a Sire 2023.0.1 patch release and will have updated to build against that instead. That should hopefully fix the issue with the outdate Sire version being pulled in for the Linux py37 build.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # BioSimSpace runtime requirements.
 
-sire ~=2023.0.0   # this matches >= 2023.0.0 && < 2023.1.0
+sire ~=2023.0.1   # this matches >= 2023.0.1 && < 2023.1.0
```
xiki-tempula, line 3: Are expecting some change for 2023.1.0?

lohedges: Potentially. 2023 is currently pre-release so the API hasn't been nailed down yet. I don't expect anything to change in a breaking way, but we don't want these pre-release builds to continue building against the full release. At that point, we'll have a 2023.1.0 release and will update the requirements accordingly.

lohedges:
Phew, everything finally passed. Will merge tomorrow along with the molecule ordering fix.

PR #402, opened 2022-11-28
Add support for autodection of disulphide bonds during parameterisation

lohedges:
This PR adds support for autodetection of disulphide bonds during parameterisation. Bonds are located using the new Sire 2023 search functionality and added as correctly formatted `LEaP` bond directives. The user can tune the tolerance to use when searching for bonds, as well as the maximum search distance. (These are helpful if the conformation is not equilibrated.) In addition, the user can pass in a list of _bonded atom pairs_ if they need to specify things explicitly. (In future, we could update this to a list of actual `Bond` objects.) Functionality has been tested by Richard Bradshaw.

The PR also adds support for passing through engine-specific keyword arguments in a few places, i.e. in the generic functions such as `BioSimSpace.MD.run`.


lohedges:
Looks like the failures are due to the wrong version of Sire being pulled in. This has been fixed in `devel`, so I'll apply this to the feature branch.

python/BioSimSpace/Parameters/Protocol/_amber.py
```diff
@@ -134,6 +199,19 @@ def __init__(self, water_model=None, leap_commands=None, property_map={}):
         # Call the base class constructor.
         super().__init__(forcefield="ff99", property_map=property_map)
 
+        # Validate the tolerance.
```
msuruzhon: Could be useful to abstract these checks to `@property` and `@property.setter`, but this is fine as is, just a general point.

python/BioSimSpace/Parameters/Protocol/_amber.py
```diff
@@ -199,18 +325,48 @@ def __init__(self, water_model=None, leap_commands=None, property_map={}):
                     raise TypeError("'leap_commands' must be a 'list' of 'str' types.")
         self._leap_commands = leap_commands
 
+        # Validate the bond records.
```
msuruzhon: There is some code duplication so could be useful to have a disulfide mixin or something like that.

python/BioSimSpace/Parameters/Protocol/_amber.py
```diff
@@ -510,18 +752,9 @@ def run(self, molecule, work_dir=None, queue=None):
                            "formal charge")
                     raise _ParameterisationError(msg)
 
-        # Create a new system and molecule group.
```
msuruzhon, line 513: Why was this needed in the past instead of just saving as a pdb?

python/BioSimSpace/Parameters/Protocol/_protocol.py
```diff
@@ -557,3 +572,167 @@ def _find_force_field(forcefield):
...
+    try:
+        disulphides = query(mol, property_map)
+    except:
+        disulphides = []
+
+    # Create a list to store the LEaP bond record strings.
+    bond_records = []
+
+    # Loop over the disulphide bonds and generate the records.
+    for bond in disulphides:
```
msuruzhon: Could this loop be made simpler with the new sire API?

python/BioSimSpace/Parameters/Protocol/_amber.py
lohedges: I believe this was there before we supported passing `Molecule` as well as `System` objects to `BSS.IO.saveMolecules()`, and also before I added things like `.toSystem()` on a `Molecule`. We also used to just write to the users working directory, so using the Sire interface directly gave a bit more flexibility with regards to file naming. (For example, you can specify the extension to use.)

python/BioSimSpace/Parameters/Protocol/_amber.py
lohedges: Good idea, I'll give this a go.

python/BioSimSpace/Parameters/_parameters.py
```diff
@@ -42,16 +42,18 @@
...
 from ..Types import Charge as _Charge
+from ..Types import Length as _Length
 from .._Utils import _try_import, _have_imported
 from .. import _Utils
 
 from ._process import Process as _Process
 from . import Protocol as _Protocol
 
-def parameterise(molecule, forcefield, water_model=None, work_dir=None, property_map={}):
+def parameterise(molecule, forcefield, water_model=None, work_dir=None, property_map={}, **kwargs):
```
msuruzhon: As far as I can see, `water_model` is not used anywhere.

python/BioSimSpace/Parameters/_parameters.py
```diff
@@ -144,19 +155,25 @@ def ff99(molecule, water_model=None, leap_commands=None, work_dir=None, property
...
                               leap_commands=leap_commands,
+                              bonds=bonds,
                               property_map=property_map)
 
     # Run the parameterisation protocol in the background and return
     # a handle to the thread.
     return _Process(molecule, protocol, work_dir=work_dir, auto_start=True)
 
-def ff99SB(molecule, water_model=None, leap_commands=None, work_dir=None, property_map={}):
+def ff99SB(molecule, tolerance=1.2, max_distance=_Length(6, "A"),
```
msuruzhon: There's a lot of similarity between at least 5 of the classes (`ff99`, `ff99SB`, `ff99SBildn`, `ff03` and `ff14SB`). Seeing that the only difference between them is basically the import statement in `tleap`, I think it might make sense to have a base `factory` class or something along these lines, that you just subclass in a trivial manner, so that the `ff99` class only calls `FF99` with some parameter (or even better, abstracted FF class with a parameter of `"protein_force_field"`). Would make maintenance much easier because it would remove a lot of duplicate code (that was my impression anyway, correct me if the duplication is needed).

test/Parameters/test_parameters.py
```diff
@@ -0,0 +1,34 @@
...
+def test_disulphide():
+    """Test parameterisation in the presence of disulphide bridges."""
+
+    # Load the example molecule.
+    molecule = BSS.IO.readMolecules("test/input/molecules/4LYT_Fixed.pdb")[0]
+
+    # Try to parameterise with ff14SB. If working, this should auto-detect
+    # disulphide bonds and add the appropriate bond records to the tLEaP input
+    # script.
+    molecule = BSS.Parameters.ff14SB(molecule).getMolecule()
```
msuruzhon: If we need to keep the separate force fields duplicate, I'd `pytest.mark.parametrize` this with every supported force field (I think I would probably do it even if it was abstracted, just to be safe).

python/BioSimSpace/Parameters/_parameters.py
lohedges: Good catch. I think the parameter shouldn't be there and would be handled by the `kwargs` since it is force field specific. I will probably need to rethink this, as it might be better to have the options handled but silently ignored by functions where they are unsupported. We need to correctly handle people calling the functions directly, or via a generic wrapper when used in a node.

test/Process/test_somd.py
```diff
@@ -37,6 +37,23 @@ def test_production(system):
     # Run the process and check that it finishes without error.
     assert run_process(system, protocol)
 
+def test_free_energy():
```
msuruzhon, line 40: Is this related to the disulfide stuff or is this just some additional unrelated test?

lohedges: Sorry, this was just an additional new test to make sure that `somd-freenrg` works, since it was broken by the Sire 2023 merge. I happened to be working on this branch at the time so added it here. It is unrelated to any of the new parameterisation functionality.

python/BioSimSpace/Parameters/_parameters.py
lohedges: These aren't classes, rather functions that call through to a generic AMBER `tLEaP` protocol that does essentially what you say, i.e. searches for the force field name in the `dat` directory (using some logic to search for deprecated force fields) and adds it to the `LEaP` script if found. (Perhaps it's the protocols that you are referring to.)

When I originally wrote this I had never parameterised a molecule before so wasn't sure how different the protocols would be for the different force fields, e.g. what options would be supported. I also wasn't sure what would and wouldn't be supported by `pdb2gmx`, which is only used as a fallback. This is why there are separate protocols for each named force field, although they all (at least the protein ones) use the same base `run` method to execute the protocol.

I think it is the _protocols_ that would likely be better abstracted in the way that you suggest, i.e. having protocols for protein and ligand / small molecule force fields, with ones for the different engines that are supported.

I'll have a go a refactoring things since I agree that this part of the code is quite messy as is.

python/BioSimSpace/Parameters/Protocol/_protocol.py
lohedges: Sorry, in what way were you thinking? It uses search functionality from 2023 to find the bonds, i.e.:

```python
# Create the search query.
query =_SireMol.Select("bonds from element S to element S")
```

I need to use the legacy API so that the property map can be respected during the search, i.e. why I use `Sire.Mol.Select` directly.

Were you thinking that the body of the loop itself could be simplified?

msuruzhon: Yes, I was under the impression there might be some new `sire` 2023 syntax to simplify the loop, but I don't know much about it so might be wrong.

msuruzhon:
Thanks for the PR, @lohedges, let me know once you are happy for me to take another look at the changes.

lohedges:
Will do. Thanks for the comments, they are really useful.

python/BioSimSpace/Parameters/Protocol/_protocol.py
chryswoods: Yes - there is 'apply' that can be used to apply a function to each result from a search (each item in a container). There may also be ways to get the names and residues via single calls. I don't think this is needed, but I think it could be a useful exercise for me to look at how this could be simplified.

lohedges: In this case there will likely only be a handful of bonds returned by the search so I don't think anything will speed things up much, if it all. That said, it would be good to know what's possible with the new API, and where it will really benefit. The current code seems pretty easy to follow as is. (It's only a few lines, after all.)

lohedges:
Okay, I've completely reworked things and now dynamically generate all of the AMBER related functions, as we do for OpenFF. I'll test things in the morning and push an update when I'm happy.

Thanks again for the thorough review. A lot of this code is effectively a first pass implementation that hasn't had been looked at in detail for a while, so it's great to have another pair of eyes on it.

Cheers

python/BioSimSpace/Parameters/Protocol/_protocol.py
chryswoods: I agree - I don't think there is much benefit in this case. It will be a useful exercise to try to apply the new api to a real world example :-)

lohedges: I was thinking that the restraint generation for ABFE might be able to be handled natively with the new trajectory analysis and searching code. It might be good to take a look at the current implementation with you at some point to see.

lohedges:
The CI is currently failing since `libxml2==2.9.12=h885dcf4_1` has been pulled. See [here](https://github.com/conda-forge/lxml-feedstock/issues/75) for details. I'll re-run once the Sire CI has finished and there is a new conda package.

lohedges:
Just a note on changes since last time:

* There are now just `AmberProtein` and `GAFF` parameterisation protocols to avoid duplication.
* The AMBER protein force field functions are generated dynamically at import time, i.e. like we do for those from OpenFF. At present these are built from a list of supported force field names, but could (when we support CMAP) be generated from querying the AmberTools installation directory directly. (This approach is nice since we don't expose functionality to the user that isn't available to them.)
* I've made `BioSimSpace.Parameters.Protocol` private, since it is only used internally at present. It can be made public if we ever decide to expose it to the user. (Originally we had thought of having a similar interface to `BioSimSpace.Process`, i.e. passing a `Molecule` and a `Protocol` to the parameterisation functions.)
* Tests for the disulphide bond functionality are now run against all compatible force fields, i.e. the AMBER protein ones.

msuruzhon:
Hi @lohedges is the PR ready for me to review again? Many thanks.

lohedges:
Yes, sorry for not pinging you, I forgot to update after resolving the CI issues.

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
```diff
@@ -1111,3 +752,195 @@ def _validate(molecule=None, tolerance=None, max_distance=None,
...
+                _open_forcefields.append(_ff)
+
+                # Create a sane function name, i.e. replace "-" and "."
+                # characters with "_".
+                _func_name = _ff.replace("-", "_")
+                _func_name = _func_name.replace(".", "_")
+
+                # Generate the function and bind it to the namespace.
+                _function = _make_openff_function(_ff)
+                setattr(_namespace, _func_name, _function)
```
msuruzhon, line 924: I think this dynamic creation has improved on the previous version of the code by a wide margin. I guess one downside of using `setattr` is that it may not play very well with IDE/Jupyter linting so these functions might not be suggested which could be confusing for the user. Not sure if we should run `_function = _make_openff_function(_ff)` explicitly for each force field to enable proper linting - it's up to you.

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
```diff
@@ -1111,3 +752,195 @@ def _validate(molecule=None, tolerance=None, max_distance=None,
     if property_map is not None:
         if not isinstance(property_map, dict):
             raise TypeError("'property_map' must be of type 'dict'")
+
+# Wrapper function to dynamically generate functions with a given name.
+# Here "name" refers to the name of a supported AMBER protein force field.
+import sys as _sys
+_namespace = _sys.modules[__name__]
+def _make_amber_protein_function(name):
```
msuruzhon, line 763: Can't we just use `functools.partial` for these decorators?

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
```diff
@@ -1111,3 +752,195 @@ def _validate(molecule=None, tolerance=None, max_distance=None,
...
+
+                # Expose the function to the user.
+                __all__.append(_func_name)
+
+                # Convert force field name to lower case and map to its function.
+                _forcefields_lower.append(_ff.lower())
+                _forcefield_dict[_ff.lower()] = getattr(_namespace, _func_name)
+
+    # Clean up redundant attributes.
+    del _base
```
msuruzhon, line 934: Why do we need these `del`s?

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
```diff
@@ -1111,3 +752,195 @@ def _validate(molecule=None, tolerance=None, max_distance=None,
...
+               values. This allows the user to refer to properties with their
+               own naming scheme, e.g. { "charge" : "my-charge" }
+
+           Returns
+           -------
+
+           molecule : :class:`Molecule <BioSimSpace._SireWrappers.Molecule>`
+               The parameterised molecule.
+        """
+        return _parameterise_amber_protein(
```
msuruzhon, line 815: Here `name` is first (which I think makes sense, because you can easily call `functools.partial` on it), while in `_parametrise_openff` `name` comes second. Should these two functions have the same parameter ordering?

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
lohedges: I tend to tidy up all temporary attributes so that they aren't exposed to the user, e.g. they won't be visible if they import the module directly. It's not necessary, but I prefer to only expose exactly what is needed and part of the module itself (or needed by it).

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
lohedges: I agree. It would be good to make both of these consistent.

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
lohedges: It works fine in Jupyter. I've not tested in an IDE, but take your point. Will see what happens.

msuruzhon: PyCharm both recognises and doesn't recognise it which is pretty funny. It shows up in the import suggestions but when you import it, it gives a red line as if the function is not there.

python/BioSimSpace/Sandpit/Exscientia/Parameters/_parameters.py
lohedges: Good point. I just reproduced the OpenFF approach, but obviously partials works in both caes :man_facepalming:

lohedges: Actually, I did it this way so that the docstrings were exposed, but I think I can use `functools.wraps` for that. Will see if I can get it to work.

lohedges: Hmm, I can't get things to work nicely both in the console and using Sphinx. I'll try a bit more, but might just keep my original approach as it avoids a whole bunch of dosctring hacking.

lohedges: Okay, I'm going to keep things as is with some minor modifications. My approach gives you the correct function signature and docstring for the console and Sphinx. Using `partial` and `update_wrapper` doesn't do this, i.e. the pre-filled parameter will still appear in the docs, which is confusing to the user.

lohedges:
@msuruzhon: Let me know if you're happy and I'll merge. I've kept things as is with regards to the decorators, since this makes the docstrings sane and works properly with Sphinx.

Cheers.

msuruzhon (approved):
Thanks @lohedges, I am happy to merge

Issue #405, opened 2022-12-16
Making HMR work with dummies


msuruzhon:
Hello,

I have just discovered that when applying HMR, the dummies are not scaled in accordance with the corresponding hydrogens. I think the expected behaviour would be for the dummies to always have the same mass as the corresponding nondummy atom. How can one make sure this is the case when calling `repartitionHydrogenMass()`? I am attaching some code for testing purposes. 

In addition, it seems that after applying HMR, the dummy gets a negative mass! (-1.00795 g/mol). That would be unexpected behaviour in any case.

Many thanks.

```python
import pytest
import BioSimSpace.Sandpit.Exscientia as BSS


@pytest.fixture
def merged_system():
    ff = "openff_unconstrained-2.0.0"
    mol0 = BSS.Parameters.parameterise("c1ccccc1", ff).getMolecule()
    mol1 = BSS.Parameters.parameterise("c1ccccc1C", ff).getMolecule()
    merged_mol = BSS.Align.merge(mol0, mol1)
    return merged_mol


def test_hmr(merged_system):
    mass0 = merged_system.getAtoms()[12]._sire_object.property("mass0")
    mass1 = merged_system.getAtoms()[12]._sire_object.property("mass1")
    assert mass0 == mass1, "This does not fail"
    merged_system.repartitionHydrogenMass(factor=3)
    BSS.IO.saveMolecules("temp", merged_system, "grotop")
    mass0 = merged_system.getAtoms()[12]._sire_object.property("mass0")
    mass1 = merged_system.getAtoms()[12]._sire_object.property("mass1")
    assert mass0 == mass1, "This will fail, but shouldn't"
```

lohedges:
Thanks for this. I hadn't thought about dummies when writing the code and agree with your suggested solution. I'll check what SOMD does too, since I believe I largely followed the implementation there, or at least tested mine against that one when first written.

lohedges:
I think the issue is that the repartitioning code generates a connectivity object to working the bonding when performing the HMR. This doesn't work properly if an atom is a dummy, since it's based on equilibrium bond radii for specific elements. I think I'll need to adjust this to use the elements from the non-dummy end state instead. Hopefully this should be easy enough to fix.

Thanks for reporting.

msuruzhon:
Thanks @lohedges, would this work if both endstates have dummies? I think that's the common use case.

lohedges:
[...]

PR #406, opened 2022-12-19
Fix dummy atom handling during hydrogen mass repartitioning

lohedges:
(This is one for the New Year.)

This PR addresses issue #405, making sure that dummy atoms are handled correctly during hydrogen mass repartitioning. Following the approach of SOMD, we repartition masses in both lambda end states by excluding dummy atoms. The masses of any dummy atoms are then assigned the repartitioned mass of the non-dummy atom in the opposite end state. This approach means that:

* Excluding dummy atoms, the mass of an alchemical end state is the same before and after repartitioning.
* Atoms perturbing to/from dummies have the same mass in both end states, as expected.

Cheers.

### DIFF
--- a/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
+++ b/python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
@@ -1350,6 +1350,16 @@ class Molecule(_SireWrapper):
 
         # Handle perturbable molecules separately.
         if self.isPerturbable():
+            # Search for dummies in both end states.
+            try:
+                dummies0 = self.search("element Xx", property_map={"element" : "element0"})
+            except:
+                dummies0 = []
+            try:
+                dummies1 = self.search("element Xx", property_map={"element" : "element1"})
+            except:
+                dummies1 = []
+
             # Repartition masses for the lambda=0 state.
             pmap = { "mass"         : "mass0",
                      "element"      : "element0",
@@ -1365,6 +1375,22 @@ class Molecule(_SireWrapper):
                    }
             self._sire_object = _SireIO.repartitionHydrogenMass(
                     self._sire_object, factor, water_options[water], pmap)
+
+            # Now replace atom dummy atom masses with the reparitioned mass
+            # from the opposite end state.
+
+            edit_mol = self._sire_object.edit()
+
+            for dummy in dummies0:
+                idx = dummy._sire_object.index()
+                mass1 = self._sire_object.atom(idx).property("mass1")
+                edit_mol = edit_mol.atom(idx).setProperty("mass0", mass1).molecule()
+            for dummy in dummies1:
+                idx = dummy._sire_object.index()
+                mass0 = self._sire_object.atom(idx).property("mass0")
+                edit_mol = edit_mol.atom(idx).setProperty("mass1", mass0).molecule()
+
+            self._sire_object = edit_mol.commit()
 
         else:
             self._sire_object = _SireIO.repartitionHydrogenMass(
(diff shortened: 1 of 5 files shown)


python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
```diff
@@ -1366,6 +1376,22 @@ def repartitionHydrogenMass(self, factor=4, water="no",
             self._sire_object = _SireIO.repartitionHydrogenMass(
                     self._sire_object, factor, water_options[water], pmap)
 
+            # Now replace atom dummy atom masses with the reparitioned mass
+            # from the opposite end state.
+
+            edit_mol = self._sire_object.edit()
```
msuruzhon, line 1382: Should we be trying to use the new `sire` API in the new PRs? So e.g. something like:

```python
c = self._sire_object.cursor()
for atom in c.atoms():
    if "du" in atom["ambertype0"]:
        atom["mass0"] = atom["mass1"]
    elif "du" in atom["ambertype1"]:
        atom["mass1"] = atom["mass0"]
self._sire_object = Molecule(c.commit())
```
Here I've used a name check for the dummies but an element check might be more appropriate.

test/Align/test_align.py
```diff
@@ -364,25 +364,54 @@ def test_hydrogen_mass_repartitioning():
...
+
+    # Get the elements in either end state.
+    elements0 = merged._sire_object.property("element0").toVector()
+    elements1 = merged._sire_object.property("element1").toVector()
+
     # Work out the initial mass of the system.
     initial_mass0 = 0
-    for mass in merged._sire_object.property("mass0").toVector():
-        initial_mass0 += mass.value()
+    for idx, mass in enumerate(merged._sire_object.property("mass0").toVector()):
```
msuruzhon, line 376: Could the new `sire` search functions simplify these? I.e. get the sum of the masses of the non-dummy atoms. I was under the impression that the new API can do that?

test/Align/test_align.py
```diff
@@ -364,25 +364,54 @@ def test_hydrogen_mass_repartitioning():
...
-        initial_mass1 += mass.value()
+    for idx, mass in enumerate(merged._sire_object.property("mass1").toVector()):
+        if elements1[idx] != dummy:
+            initial_mass1 += mass.value()
 
     # Repartition the hydrogen mass.
     merged.repartitionHydrogenMass()
 
+    # Lists to store the mass of dummy atoms in the two end states.
+    dummy_masses0 = []
```
msuruzhon, line 388: Similarly, `sire` might have a function that will give you the dummy and non-dummy lists as a one-liner?

msuruzhon:
Thanks @lohedges and happy new year! The PR looks good but I was wondering whether some of the operations could be simplified with the new `sire` API. Apart from that it looks good to me.

python/BioSimSpace/Sandpit/Exscientia/_SireWrappers/_molecule.py
lohedges: It could, but in the existing Sire code (not the PR that is yet to be merged) you can't use a property map to index in this way, so the property naming _could_ be wrong. That said, I agree that it would be good (where appropriate) to start using the new API in PRs so that the code gets slowly ported over. I'd prefer to do it in self-contained functions, or with new functionality, so that we don't end up with a weird mixed-API that might be confusing for developers to follow. (In this case, it would be easy to replace the whole thing.)

test/Align/test_align.py
lohedges: I think so, but believe that this might need the latest 2023 updates that are yet to be merged. Here I just quickly edited the existing unit test, rather than re-writing the whole thing for the new API.

test/Align/test_align.py
lohedges: Yes, you can now do things like `mol["not element Xx"].property("mass")`, but this doesn't seem to work for version of Sire used for this PR. Also, extracting elements for a non-standard property name (as in this case) doesn't seem to work at present.

msuruzhon (approved):
Thanks @lohedges it looks like there are still some issues to iron out with the new `sire` so I am happy to merge as is.

PR #408, opened 2023-01-09
Updates following Sire 2023 API changes

lohedges:
This PR addresses API updates following the Sire 2023.X PR [here](https://github.com/michellab/Sire/pull/419). Specifically:

* Several search strings have been updated so that they are performed at the correct view level.
* The code has been updated to correctly handle `sire.legacy.Maths.Vector`, the elements of which now have units of length (applied in the Python layer).
* Fix the extraction of coordinates from residue objects, since this is now supported natively by Sire.
* Due to the requirements of our dependencies, and Sire 2023, our license has now been updated to GPLv3 or later.

I have also taken the opportunity to blacken the _entire_ codebase, i.e. source code, notebooks, helper scripts, etc. Following this change I noticed that black applies at least four different (inconsistent) formatting styles to docstrings, so manually fixed these to apply a single style throughout. (I assume that this is done to minimise diffs, assuming certain requirements are met.) I have also applied an early `pytest --black` test to the CI to catch formatting errors. I will think about applying some linting checks at a later date, although this might be tricky due to the way BioSimSpace dynamically generates many attributes at import time.

Note that I currently need to pin to `alchemlyb` version < 2, since `alchemlyb.estimators.AutoMBAR`, which is used within the Exscientia sandpit, has been deprecated. This can be removed once we have fully tested and updated against version 2.0. (I believe @annamherz has tested this, but I haven't merged in edits yet since they aren't based off `devel`. While the edits are simple, so could easily be applied to the sandpit, `devel` still uses the legacy analysis scripts which require `pymbar` < 4.)

(Apologies that the diffs for this PR are large. In reality, very little has changed, so as long as your internal test suite passes, then all should be okay.)

Extra things that I might be able to get done by 2023.1:

* Update the FEP analysis code from @annamherz's [branch](https://github.com/michellab/BioSimSpace/tree/feature-analysis). There are a few things to iron out here, since the branch now contains duplicate functionality in different parts of BioSimSpace, i.e. overlap matrix plotting.
* Review / merge @fjclark's SOMD ABFE [PR](https://github.com/Exscientia/BioSimSpace/pull/19) into `Exscientia/devel` (and potentially into `michellab/devel`.) 
* Go through CCPBioSim workshops [here](https://github.com/CCPBioSim/biosimspace-workshop) and [here](https://github.com/CCPBioSim/biosimspace-advanced-simulation) to make sure that they work with the new code. Update the respective READMEs with installation instructions for _specific_ Sire and BioSimSpace versions so that they continue to work.

Cheers.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -77,33 +77,28 @@ jobs:
         run: mkdir ${{ github.workspace }}/build
 #
       - name: Build Conda package using mamba build
-        run: conda mambabuild --croot ${{ github.workspace }}/build -c conda-forge -c michellab/label/dev ${{ github.workspace }}/BioSimSpace/recipes/biosimspace
-        if: matrix.platform.name != 'windows' || matrix.python-version != '3.7'
+        run: conda mambabuild -c conda-forge -c michellab/label/dev ${{ github.workspace }}/BioSimSpace/recipes/biosimspace
 #
-      - name: Build Conda package using conda build
-        run: conda build --croot ${{ github.workspace }}/build -c conda-forge -c michellab/label/dev ${{ github.workspace }}/BioSimSpace/recipes/biosimspace
-        if: matrix.platform.name == 'windows' && matrix.python-version == '3.7'
+#      - name: Collect failed artifacts
+#        run:
+#          python ${{ github.workspace }}/BioSimSpace/actions/collect_failed.py
+#        env:
+#          BUILD_DIR: ${{ github.workspace }}/build
+#        if: ${{ failure() }}
 #
-      - name: Collect failed artifacts
-        run:
-          python ${{ github.workspace }}/BioSimSpace/actions/collect_failed.py
-        env:
-          BUILD_DIR: ${{ github.workspace }}/build
-        if: ${{ failure() }}
+#      - uses: actions/upload-artifact@v3
+#        with:
+#          name: build-${{ matrix.platform.name }}-${{matrix.python-version}}-${{ github.sha }}
+#          path: ${{ github.workspace }}/build/*/biosimspace-*.tar.bz2
+#          retention-days: 10
+#        if: ${{ success() }}
 #
-      - uses: actions/upload-artifact@v3
-        with:
-          name: build-${{ matrix.platform.name }}-${{matrix.python-version}}-${{ github.sha }}
-          path: ${{ github.workspace }}/build/*/biosimspace-*.tar.bz2
-          retention-days: 10
-        if: ${{ success() }}
-#
-      - uses: actions/upload-artifact@v3
-        with:
-          name: failed-${{ matrix.platform.name }}-${{matrix.python-version}}-${{ github.sha }}
-          path: ${{ github.workspace }}/build/failed.tar.bz2
-          retention-days: 1
-        if: ${{ failure() }}
+#      - uses: actions/upload-artifact@v3
+#        with:
+#          name: failed-${{ matrix.platform.name }}-${{matrix.python-version}}-${{ github.sha }}
+#          path: ${{ github.workspace }}/build/failed.tar.bz2
+#          retention-days: 1
+#        if: ${{ failure() }}
 #
       - name: Upload Conda package
         run: python ${{ github.workspace }}/BioSimSpace/actions/upload_package.py
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -65,7 +65,10 @@ jobs:
         if: github.event_name == 'pull_request'
 #
       - name: Setup Conda
-        run: mamba install -y -c conda-forge boa anaconda-client packaging=21 pip-requirements-parser
+        run: mamba install -y -c conda-forge boa anaconda-client packaging=21 pip-requirements-parser pytest-black
+#
+      - name: Check formatting
+        run: PYTHONPATH=${{ github.workspace }}/BioSimSpace/python python -m pytest -svvv --color=yes --black ${{ github.workspace }}/BioSimSpace/python
 #
       - name: Update Conda recipe
         run: python ${{ github.workspace }}/BioSimSpace/actions/update_recipe.py
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -65,10 +65,7 @@ jobs:
         if: github.event_name == 'pull_request'
 #
       - name: Setup Conda
-        run: mamba install -y -c conda-forge boa anaconda-client packaging=21 pip-requirements-parser pytest-black
-#
-      - name: Check formatting
-        run: PYTHONPATH=${{ github.workspace }}/BioSimSpace/python python -m pytest -svvv --color=yes --black ${{ github.workspace }}/BioSimSpace/python
+        run: mamba install -y -c conda-forge boa anaconda-client packaging=21 pip-requirements-parser
 #
       - name: Update Conda recipe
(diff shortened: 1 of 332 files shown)


lohedges:
I just need to fix the Sire 2023.0.3 build, as some of the BioSimSpace requirements have been removed for the Windows platform, e.g. `mdtraj` and `mdanalysis`. This means that it is impossible to build BioSimSpace on top of Sire for this platform.

msuruzhon:
Thanks @lohedges should we also think about automatic docstring formatter, as manually changing the output from `black` would still mean that we can't just set and forget. I am trying to solve the merge "conflicts" (just formatting) with our version and the easiest thing for me would be to just apply an autoformatter to all the files. Right now I get conflicts in pretty much all of the docstrings and I am not sure what the most efficient way of dealing with this is.

lohedges:
Yes, I can look into that. I only adjusted since, in some case, it actually affected the rendering in Sphinx and the Python help, so I'd prefer to have things consistent at the point at which black was first applied, i.e. now.
Do you have any recommendations for an autoformatter to use for docstrings, e.g. what do you use internally?

What branch are you trying to merge with at your end? I checked your `devel` beforehand, i.e. the source of reference for this PR, and it is only 11 commits ahead of ours, and did not appear to be blackened. Maybe you have done this on a feature branch. If so, we need to think about the point of reference for PRs in future, i.e. the synchronisation is currently `devel` to `devel`.

lohedges:
I had also (wrongly) assumed that `black` would apply a consistent docstring autoformatting of its own. I've taken a look at [this](https://github.com/DanielNoord/pydocstringformatter) and it seems to match the style that I've used if run with:

```
pydocstringformatter --no-split-summary-body python/BioSimSpace 
```

However, it appears to incorrectly apply formatting to some Sphinx specific docstrings, which might break things. I'll have a play around to see if I can get it to work.

lohedges:
Yes, it completely breaks Sphinx :clap:

lohedges:
Okay, I can fix it afterwards with `sed`. I'll send you a script so that you can apply things your end.

lohedges:
Right, the following seems to work:

1) Autoformat the docstrings.

```
pydocstringformatter --no-split-summary-body --style=numpydoc --write python/BioSimSpace
```

2) Fix all of the Sphinx docstrings.

```
sed -i.bak -e '/currentmodule/ s=.$==' $(grep -r currentmodule python/BioSimSpace | awk -F ':' '{print $1}')
rm $(find . -name "*.bak")  
```

Let me know if you're happy with this and I'll apply, then push the changes. (Note that I've re-run `black` after doing the above and it's happy with everything, i.e. we don't want to get into a situation where we create an infinite autoformatting loop.)

msuruzhon:
Thanks @lohedges, I think my issue is that I applied `black` in [this](https://github.com/Exscientia/BioSimSpace/tree/feature_merge_sire_2023) merge branch before I merged `feature-2023_0`, so I got all of these conflicts. Thanks for the script, I will give it a try!

msuruzhon:
Apologies for the silly question @lohedges, but what would `currentmodule` look like? Could you give an example? Maybe on how to apply it on the whole `python/BioSimSpace`?

lohedges:
This is one of the specific docstrings that are processed by Sphinx in order to generate the website, e.g. from [here](https://github.com/michellab/BioSimSpace/blob/63b14996a0583b44822204f3b8e2a246b8dcd986/python/BioSimSpace/Align/__init__.py#L27):

```python
"""
.. currentmodule:: BioSimSpace.Align

...
```

This issue is that, by default, `pydocstringformatter` applies periods at the end of docstring sentences. This is a good thing, but then Sphinx is unable to parse things correctly and I end up with missing documentation. My `sed` command just deletes the final character for any line containing `currentmodule` within the `python/BioSimSpace` directory.

msuruzhon:
Ah okay, so this command is ready to use within the main directory then? (I thought `currentmodule` was a placeholder for something) Because in this case when I run 

```
sed -i.bak -e '/currentmodule/ s=.$==' $(grep -r currentmodule python/BioSimSpace | awk -F ':' '{print $1}')
rm $(find . -name "*.bak")  
```

I get:

```sed: Binary: No such file or directory```

Could this be a macOS thing?

**Comment by lohedges** [comment]
Yes, this is probably a macOS thing, since it uses BSD rather than GNU `sed`. I thought I'd made the command POSIX compliant, but apparently not. I'll see if I can adjust it for you.

**Comment by lohedges** [comment]
It is actually the `grep` output that is different on macOS, since the formatting for binary matches is different. This works for me:

```
sed -i.bak -e '/currentmodule/ s=.$==' $(grep -r currentmodule python/BioSimSpace | grep '.py:' | awk -F ':' '{print $1}')
```

**Comment by msuruzhon** [comment]
Amazing, that works for me too, thanks a lot @lohedges!

**Comment by msuruzhon** [comment]
Hi again @lohedges, I think I got most of our tests to pass now, but I am having an issue importing `angstrom2` from `sire`. Do you know what the new way of doing it is? I am looking at the following code snippet:

```
force_constant = force_constant.to(
    _SireUnits.kcal_per_mol / _SireUnits.angstrom2
)
```

chryswoods:
We didn't expose all of the SireUnits units when we were developing 2023.X. I am fixing this now - the 2023.1.0 release will expose `sire.units.angstrom2`, which will be `_SireUnits.angstrom2` in BioSimSpace.

msuruzhon:
Thanks @chryswoods let me know when that's ready to test and I'll give it a try.

lohedges:
For now you can just use `sire.legacy.Units`. I've needed to do this throughout and plan to update (where appropriate) some time after the 2023.1.0 release.

lohedges:
Ah yes, I see that I've missed using `sire.legacy.Units` in a few places. Will fix this now.

lohedges:
Okay, hopefully that has fixed things. (Note that I've also pushed the docstring autoformatting changes using the approach described above.) Apologies for missing these. I had updated the files that were triggered by failures of the existing tests, but had missed a few places that weren't covered. I guess your internal test suite is more extensive.

msuruzhon:
Thanks, these were just our `Sandpit` tests - I think I messed up the merge a bit actually but it seems it was for the better! Once I get all tests to pass, I will test internally as well.

lohedges:
Ah, they must have been ones that ran conditionally, perhaps the ABFE restraint ones. I'll try to fix this so that everything is run on PR like this. We do already run your tests, but not using the same CI runner script as you.

annamherz:
> Note that I currently need to pin to `alchemlyb` version < 2, since `alchemlyb.estimators.AutoMBAR`, which is used within the Exscientia sandpit, has been deprecated. This can be removed once we have fully tested and updated against version 2.0. (I believe @annamherz has tested this, but I haven't merged in edits yet since they aren't based off `devel`. While the edits are simple, so could easily be applied to the sandpit, `devel` still uses the legacy analysis scripts which require `pymbar` < 4.)

yep, the edits for it are in the `feature-analysis` 😊

msuruzhon (approved):
So I also tested it with our internal code - it seems that the only bit which needed changing was a place in the code where I was converting a `Vector` to a `np.ndarray` - this used to convert directly, now it gives units as well so it has to be extracted with `vector.x().value(), ...`. But everything else seems fine.

lohedges:
Fantastic, thanks for confirming. I'm working on a second set of edits [here](https://github.com/michellab/BioSimSpace/tree/feature-2023_1), but these mostly relate to website updates and wrapping the new Sire load functionality. I'll let you know when I'm done. This is likely to be the final update prior to 2023.1.0.

PR #416, opened 2023-01-31
Sync the two branches

xiki-tempula:
Given that there are some formatting changes, it might be good to sync the branch before further conflict happens.

Issue #413, opened 2023-01-24
`_toRegularMolecule` failing with `convert_amber_dummies=True`


msuruzhon:
Hi,

I have observed a strange bug when executing the following code:

```python
import BioSimSpace as BSS

# Generate the two ligands
ff = "openff_unconstrained-2.0.0"
mol0 = BSS.Parameters.parameterise("c1ccccc1", ff).getMolecule()
mol1 = BSS.Parameters.parameterise("c1ccccc1C", ff).getMolecule()

# Generate a merged mol
mapping = BSS.Align.matchAtoms(mol0, mol1, complete_rings_only=True)
mapping.pop(9)  # So that we have a dummy in both endstates
merged_mol = BSS.Align.merge(mol0, mol1, mapping=mapping)

# This works
merged_mol._toRegularMolecule(is_lambda1=False, convert_amber_dummies=True)

# This fails?
merged_mol._toRegularMolecule(is_lambda1=True, convert_amber_dummies=True)
```
with the following traceback:

```
line 1699, in _toRegularMolecule
    mol.atom(dummy.index())
Boost.Python.ArgumentError: Python argument types in
    Selector_Atom_.index(Selector_Atom_)
did not match C++ signature:
    index(SireMol::Selector<SireMol::Atom> {lvalue}, int i)
```

It seems that the dummy atom somehow doesn't have an `.index()` attribute?

Any insights will be appreciated. Many thanks.

lohedges:
Hmm, in the first case if I print the type of the search result in the loop I see:

```python
Atom( H4x:13  [  -1.81,    2.20,    0.01] ) <class 'sire.legacy.Mol._Mol.Atom'>
Atom( H6x:14  [   2.57,   -1.08,    0.08] ) <class 'sire.legacy.Mol._Mol.Atom'>
Atom( H7x:15  [   2.47,   -0.04,   -1.36] ) <class 'sire.legacy.Mol._Mol.Atom'>
Atom( H8x:16  [   2.71,    0.67,    0.33] ) <class 'sire.legacy.Mol._Mol.Atom'>
```

In the second case I get:

```python
Atom( H4x:10  [   2.15,   -1.17,    0.23] )
) <class 'sire.legacy.Mol._Mol.Selector_Atom_'>
```

Somehow we're ending up with a `Selector_Atom` rather than a regular `Atom`. I'll try to figure out what's gone wrong.

lohedges:
The following diff seems to work:

```
[...]

PR #417, opened 2023-02-06
Sync with OpenBioSim/devel [closes #416]

lohedges:
This PR pulls in updates from [OpenBioSim/devel](https://github.com/OpenBioSim/biosimspace) to synchronise the two repositories for the 2023.1.0 release of BioSimspace. See the [CHANGELOG](https://biosimspace.openbiosim.org/changelog.html#january-31-2023) for details of what's changed.

Note that this PR supersedes and closes #416. Changes and updates to the Exscientia sandpit are as follows:

* Autoformat with black 23.
* Add basic support for file caching, i.e. we check to see if the system UID and version has changed on write and copy the existing file (if it sill exists and hasn't changed) if they are the same. Note that this _only_ works if you use `BioSimSpace.IO.saveMolecules` to write files so I've updated `write_system` in your AMBER and GROMACS process code to use this. (I will work on improving performance, but this is a starting point. In the current code, versions of systems stored by different processes will differ because of the way systems are copied, so I'll need to think about this too.)
* Remove use of `mklinkto` in tests since it doesn't work across file systems. Instead I've copied the file into the processes temporary directory at the start of the test. This will work assuming write permissions on the path.
* Change the fix for duplicate `nsteps = 0` config entries in GROMACS tests so that the option is passed in as `extra_options={"nsteps": 0}`.  This avoids the need to modify the config, since the `extra_options` overwrite the defaults.
* Guard tests against missing packages so that they run on _all_ platforms that we support. (Need to guard against `ambertools`, `gromacs`, `mdtraj`, `mdanalysis`, `openff`.)
* Fix scoping of fixtures so that test systems/molecules aren't repeatedly re-created during testing.
* Updated all tests to use remote files where possible to greatly reduce repository size. For future PRs you can upload files to the thread and use the permalink in the test. We can then upload these files to our website and update the links on merge.

Many thanks.


xiki-tempula:
> Add basic support for file caching, i.e. we check to see if the system UID and version has changed on write and copy the existing file (if it sill exists and hasn't changed) if they are the same. Note that this only works if you use BioSimSpace.IO.saveMolecules to write files so I've updated write_system in your AMBER and GROMACS process code to use this. (I will work on improving performance, but this is a starting point. In the current code, versions of systems stored by different processes will differ because of the way systems are copied, so I'll need to think about this too.)

Is the system UID for topology for the whole system? I think the major use case that we have is that the writing of topology is slow so we want to avoid repeated writing the same topology, but the coordinate, velocity, box dimension could differ.

lohedges:
> Is the system UID for topology for the whole system? 

Yes, this is what was discussed in the meeting when you were away.

>  ...coordinate, velocity, box dimension could differ.

Okay, if this is the case then I'll need a rethink. I guess it might be possible to add something in Sire that could exclude certain properties when generating a checksum, although this might be complicated in practice and potentially more time consuming than writing the topology in the first place. Will chat with @chryswoods to see if he has any suggestions.

xiki-tempula:
@lohedges I know that Sire stored such property at atom level so it might be difficult. OpenMMTools has a object called sampler_state which only holds the coordinate, velocity, box dimension (with no topology information in it), which I think is quite convenient.

lohedges:
Thanks. Yes, something like that would be good. The main issue is that Sire is so flexible, i.e. pretty much anything can be edited or be added as a property, so checking that an entire `System` is the same is challenging. This is why the UID and version makes things quick. We could group properties together somehow and checksum based on those, i.e. topology related ones, state related ones, but this would require knowing exactly what properties are required for a specific format and protocol. (Which is possible, but might be fiddly as things change or are updated.) 

For simple cases where we you are working with a fixed system, then we could probably do something like just checking the first molecule to see if anything has changed, e.g. just in Python we could create a dictionary that hashes each of the system properties and those in the first molecule, e.g.:

```python
import BioSimSpace as BSS

excluded_props = ["coordinates", "space"]

prop_hash = {}

url = BSS.tutorialUrl()
files = BSS.IO.expand(url, ["ala.crd", "ala.top"], ".bz2")
system = BSS.IO.readMolecules(files)

for prop in system._sire_object.propertyKeys():
    if prop not in excluded_props:
        prop_hash[prop] = hash(prop)

for prop in system[0]._sire_object.propertyKeys():
    if prop not in excluded_props:
        prop_hash[prop] = hash(prop)

for prop, _hash in prop_hash.items():
    print(prop, _hash)
```

Gives:

```
Downloading from 'https://biosimspace.openbiosim.org/m/ala.crd.bz2'...
Unzipping '/tmp/tmp9_64sbqs/ala.crd.bz2'...
Downloading from 'https://biosimspace.openbiosim.org/m/ala.top.bz2'...
Unzipping '/tmp/tmp9_64sbqs/ala.top.bz2'...
filenames 3022442931254130624
fileformat 5312110957873926959
ambertype -4632611405162013524
bond -7216836253032058868
gb_radius_set -7114445824508384328
dihedral 3192620588422740254
gb_screening -8072084404698643768
mass 602409594107741342
element -4453701148717323862
angle -4294095326391043727
gb_radii 6451370362249813749
parameters 5767477298835769504
intrascale -6008015028032039997
forcefield -8340658508579024993
atomtype 8706876941176840703
connectivity -4625670802376471359
LJ -2764781884618937059
charge 1074835261070421639
treechain -8994582374208645934
improper -5691352357823464926
```

These could then be checked prior to writing the system. Would need to think how to apply something like this more generally, e.g. so that we could do full checks when necessary, but simple checks like above when we know that the system is the same (other than `coordinates`, `velocity`, and `space`), which would be the case if you chain protocols like minimisation, equilibration, etc., without doing any edits in between.

lohedges:
Having chatted with @chryswoods it seems like we should be able to use the fact that all properties are versioned. With the new API, it's very easy (and quick) to check the versions of any properties in the system, e.g. the following would give a list containing the version of the coordinates property for all molecules.

```python

import BioSimSpace as BSS
from sire.legacy.Mol import SelectorMol

url = BSS.tutorialUrl()
files = BSS.IO.expand(url, ["ala.crd", "ala.top"], ".bz2")
system = BSS.IO.readMolecules(files)

versions = SelectorMol(s._sire_object.molecules()).apply("version", "coordinates")
```

The next step is to get each parser to expose what property keys it uses, e.g. for the`GroTop` parser, there could be a function `GroTop.property_keys()` that returns a list of the property names, e.g. `["bond", "angle", ...]`. Then, for a system with a given UID, we could check whether we need to re-write a file to a given format by checking if the property versions are the same as the last time it was written to that format, e.g.:

```python
# Dictionary to store property keys and the version for each molecule.
prop_versions = {}

# Below 'mols' is a sire System, or a SelectorMol object.

for prop in GroTop.property_keys():
    try:
        prop_versions[prop] = mols.apply("version", prop)
    except:
        prop_versions[prop] = None

# Check keys match against previous prop_versions and for each key
# check that the value is the same.
```

With this approach we have parser-specific checksumming. It would probably be easy enough to monkey-patch in the `Parser.property_keys` functions in the Python layer (or with BioSimSpace) too.

xiki-tempula:
This plan looks promising

xiki-tempula:
This is PR ready?

lohedges:
I have some minor edits on `openbiosim/devel` that addresses the issue that @msuruzhon was seeing in #413 that I was planning on merging across, along with some fixes to standardise the creation of working directories, since I found a minor issue with relative paths following the switch to using `IO.saveMolecules` throughout to take advantage of file caching. I'll let you know when those are ready.

Cheers.

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
```diff
@@ -342,8 +343,8 @@ def toString(self, engine="Gromacs"):
     def correction(self):
         """Give the free energy of removing the restraint."""
         if self._restraint_type == "boresch":
-            K = (
-                _k_boltz * (_kcal_per_mol / _kelvin) / (_kj_per_mol / _kelvin)
+            K = _k_boltz.value() * (
```
xiki-tempula, line 346: Is this intentional?

lohedges:
```python
            K = _k_boltz.value() * (
                _kcal_per_mol / _kj_per_mol
            )  # Gas constant in kJ/mol/K
```

The `_kelvin` terms cancel. With Sire 2023.1.0 `_kboltz` now has units, so you need to call `.value()` to get it as a float.

xiki-tempula: Thanks, but I was importing from sire legacy here so it is a bit worrying that these things got changed.

lohedges: I'm not sure if this is a change in the Python layer, not the C++ object itself. (Will check.) The code is versioned against Sire so it will only build against `sire~=2023.1.0`. I guess you are just copying the branch locally and testing in an existing environment.

xiki-tempula: Thanks for the explanation. Sorry, I was under the impression that Sire changes but Sire.legacy retains the old behaviour forever. So this means that one should assume that Sire.legacy is updating as well?

lohedges: Yes, the objects in both APIs are identical, so it's a change in the C++ layer:

```python
from sire.units import k_boltz
type(k_boltz)
sire.legacy.Units._Units.GeneralUnit
```

```python
from sire.legacy.Units import k_boltz
type(k_boltz)
sire.legacy.Units._Units.GeneralUnit
```

From the changelog [here](https://sire.openbiosim.org/changelog.html#id1) all constants now are exposed with units, rather than being raw floats.

xiki-tempula: I see. Thanks.

lohedges: > Thanks for the explanation. Sorry, I was under the impression that Sire changes but Sire.legacy retains the old behaviour forever. So this means that one should assume that Sire.legacy is updating as well?

Sorry, I responded before seeing this question. Yes, your assumption is what I thought initially when 2023 was first in development, and there _are_ different legacy and non legacy objects, e.g. `System`. I think the non-legacy part is simply the new Python wrapped types and utility functions from [here](https://github.com/OpenBioSim/sire/tree/devel/src/sire). These still use the underlying _legacy_ C++ objects behind the scenes, which have needed to be updated in certain cases to fix bugs and make things self-consistent, e.g. with regards to units.

xiki-tempula: Thanks. I will assume that the unit related things are changing in the sire.legacy.

lohedges: No problem. We always fix your code / tests when introducing changes, but I'll try to do a better job at clarifying what has changed and why during a PR so that there are no surprises.

chryswoods: Sorry - `sire.legacy` tries to retain as much old behaviour as possible, but this won't be forever, and there are some small things that will change. As much as possible I am trying to minimise any changes, e.g. by having separate legacy and non-legacy objects, like `System`. The small things that change are to fix confusing or ambiguous behaviour. In this case, `k_boltz` was one of a small number of physical constants that didn't have units. This created issues when using the new `GeneralUnit` class, and so had to be updated.

As things have evolved I have realised that `legacy` is the wrong name, but I won't change it now ;-). All the C++ wrappers and C++ code is in legacy, and any new C++ code will also be added to legacy. The "new" API is fully python, and sits in the `sire` namespace. As we move forwards, the backwards-compatibility guarantees will be applied based on what is in the "new" API, and how it is described in the tutorials.

We are using the unit tests, plus larger integration tests in the [sire_bigtests](https://github.com/OpenBioSim/sire_bigtests) repo to minimise the risk of breakage caused by updates. Assuming good unit tests, we should (hopefully) be able to submit pull requests to fix any changes on your branch before you notice that there is a problem ;-)

xiki-tempula: Thanks. I do understand that maintaining legacy and current version is quite difficult

xiki-tempula (approved):
This PR passes our internal test suite

python/BioSimSpace/Sandpit/Exscientia/FreeEnergy/_restraint.py
chryswoods: Sire hit 18 years old last weekend, so there is a lot of legacy ;-)

lohedges: Now is legally able to handle OH groups.

xiki-tempula:
@lohedges Is the files test/Sandpit/Exscientia/input/protein_ligand/ in the url?

lohedges:
Apologies, I've uploaded them but didn't spot the bit where you are loading them directly with `MDAnalysis`. Presumably this is where this is failing. Until we have support for remote trajectory reading feel free to add them back into the test input file directory.

The `readMolecules` parts should work fine, but this is currently failing:

```
    def getTrajectory(self, format="mdanalysis"):
        return mda.Universe(
            "test/Sandpit/Exscientia/input/protein_ligand/complex.tpr",
            "test/Sandpit/Exscientia/input/protein_ligand/traj.xtc",
        )
```

(Those files are on the website, but there is no want to read via the URL.)

xiki-tempula:
Cannot it be loaded use a api similar to
```
    def getTrajectory(self, format="mdanalysis"):
        return mda.Universe(
            f"{url}/protein_ligand/complex.tpr",
            f"{url}/protein_ligand/traj.xtc",
        )
```
?

lohedges:
Only if the `MDAnalysis` developers are willing to support this, right? I could do something in the `BioSimSpace.Trjajectory` layer, but this isn't a priority right now. (Then you could load it with BioSimSpace and convert it to `MDAnalysis` format.)

xiki-tempula:
Oh, I see now. Maybe the easier way would be to have a wrapper?
```
    def getTrajectory(self, format="mdanalysis"):
        return mda.Universe(
            BSS.IO.url(f"{url}/protein_ligand/complex.tpr"),
            BSS.IO.url(f"{url}/protein_ligand/complex.tpr"),
        )
```
Which should download the file to temporary dir and return the string path?

lohedges:
Yes, that's essentially what the `sire.load` code does, i.e. detects if the file has a URL in it, then downloads to a temporary directory. I can look at re-using this for trajectories when I get a chance. For now, we can just re-instate the files that you need. (At present the trajectory code is mostly used internally using files written by processes, so it never really needs to read from a URL.)

xiki-tempula:
Great. Thanks.

lohedges:
This was very easy to add and I've had done so on my OpenBioSim feature branch. Next time I synchronise back you should be able to test at your end.

Cheers.

PR #420, opened 2023-02-24
Sync with openbiosim/devel

lohedges:
This PR pulls in updates from the [devel](https://github.com/OpenBioSim/biosimspace) channel at `openbiosim`. See the [CHANGELOG](https://biosimspace.openbiosim.org/changelog.html#feb-24-2023) for details.

Changes that affect the Exscientia sandpit are as follows:

* Added support for loading remote trajectory files. I have updated your tests accordingly and removed the files from the repository. (The ones from the website are now used.)
* The code has been refactored to use a unified [WorkDir](https://biosimspace.openbiosim.org/api/generated/BioSimSpace._Utils.WorkDir.html#BioSimSpace._Utils.WorkDir) utility class for creating and handling working directories throughout the code base. This behaves as a string, e.g. so you can still use it in, e.g., format strings to create input file names. However, if you have any utility functions that expect a `work_dir` to be passed in as a string (rather than the new type) then you'll need to update these to pass in `str(work_dir)` rather than `work_dir`. (I have already done this throughout the codebase, but you might have internal tests/tools that need updating.)
* Fixed some issues with mixtures of absolute and relative paths following the switch to using [BioSimSpace.IO.saveMolecules](https://biosimspace.openbiosim.org/api/generated/BioSimSpace.IO.saveMolecules.html#BioSimSpace.IO.saveMolecules) throughout. This affected the use of OpenForceField parameterisation functions when a`work_dir` was specified, i.e. it wasn't run in a temporary directory.
* Added property-based system comparison with the [System.isSame](https://biosimspace.openbiosim.org/api/generated/BioSimSpace._SireWrappers.System.html#BioSimSpace._SireWrappers.System.isSame) method. This allows you to compare two systems based on a set of common properties, excluding those that are irrelevant, e.g. coordinates, velocities, etc., if you just want to check whether the topology is the same. This is now used by our internal file cache, which significantly improves performance. To expose the functionality in your Sandpit code, you can pass the `excluded_properties` keyword argument to the [BioSimSpace.IO.saveMolecules](https://biosimspace.openbiosim.org/api/generated/BioSimSpace.IO.saveMolecules.html#BioSimSpace.IO.saveMolecules) function to tell it what properties to ignore when comparing systems. (I'd be interested to know how much this improves performance.) Note that the system comparisons still assume the same UID, so it won't work for two identical systems loaded from file at a different time. (The check is only fast because it uses the UID and property version number, which are quick to validate. A full property-by-property check would be very slow, i.e. certainly slower than writing the system.)
* Added the initial [BioSimSpace.Convert](https://biosimspace.openbiosim.org/api/index_Convert.html) sub-package, which wraps around the new `sire.convert` functionality. This allows conversion between BioSimSpace, Sire, and RDKit objects, as well as conversion to BioSimSpace molecules from SMILES strings. I've updated the internal code to use the direct RDKit conversion in several places, rather than going via file. (Conversion via file is still performed if the direct conversion fails, and the user can force file conversion when direct conversion is known to be problematic.)

Cheers.


xiki-tempula:
I think our internal test is certainly breaking because of 

> The code has been refactored to use a unified [WorkDir](https://biosimspace.openbiosim.org/api/generated/BioSimSpace._Utils.WorkDir.html#BioSimSpace._Utils.WorkDir) utility class for creating and handling working directories throughout the code base. This behaves as a string, e.g. so you can still use it in, e.g., format strings to create input file names. However, if you have any utility functions that expect a work_dir to be passed in as a string (rather than the new type) then you'll need to update these to pass in str(work_dir) rather than work_dir. (I have already done this throughout the codebase, but you might have internal tests/tools that need updating.)

Our test failed with `process.setArg(process._work_dir, True)`

lohedges:
Okay, you should just need to change that to one of the following:

```python
process.setArg(process.workDir(), True)
process.setArg(str(process._work_dir), True)
```

